PR #4, opened 2019-06-07
Infer ARCH from emulators' target names

jrtc27:
This allows things like make c_emulator/cheri_riscv_sim_RV{32,64} to work, and they can be built in parallel.

The plain sail-riscv should probably also get this if approved and somebody has the time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 # Simplified makefile for CHERI.
+# Default architecture to build for all and non-namespaced targets
 ARCH ?= RV64
 ifeq ($(ARCH),32)
   override ARCH := RV32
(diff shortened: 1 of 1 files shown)


jrtc27:
I'll work on this a bit more. It's not very resilient against doing stupid things like `make c_emulator/cheri_riscv_sim_RV17`, and even `make generated_definitions/c/riscv_model_RV32.c` on its own (no ARCH and not built implicitly from the simulator's rule) will build an RV64 version.

rmn30:
Maybe we should just have a target `c_sim` and do make make ARCH=32/64 c_sim ?

jrtc27:
> Maybe we should just have a target `c_sim` and do make make ARCH=32/64 c_sim ?

You still can't build them in parallel that way.

PR #17, opened 2020-04-30
Code cleanup in preparation for sentries

jrtc27:
These should all be NFCs. Further work will be needed to support putting sentries in xEPCC and not immediately untagging them due to legalisation, and likely with changes to sail-riscv too as currently prepare_xret_target's return value is masked, whereas it would be helpful for the function itself to have control over/knowledge of the masking in order to determine whether a sealed cap is being modified. We also need to stop legalising CSR/SCRs at write-time and move to the correct read-time legalisation, which already can yield observable differences in edge cases even without CHERI.

### DIFF
--- a/src/cheri_prelude_128.sail
+++ b/src/cheri_prelude_128.sail
@@ -36,10 +36,12 @@
 type cap_size : Int = 16
 let cap_size = sizeof(cap_size)
 let log2_cap_size = 4
-type CapBits = bits(8*cap_size)
+type CapBits = bits(8 * cap_size)
 /* width of otype field in bits */
-let otype_width = 18
-let uperms_width = 4
+type otype_width : Int = 18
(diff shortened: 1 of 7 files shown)


pmundkur:
lgtm

src/cheri_prelude_128.sail
```diff
@@ -75,10 +77,10 @@ struct Capability = {
   internal_e             : bool    ,
   E                      : bits(6) ,
   sealed                 : bool    ,
```
arichardson, line 79: Do we still need the sealed boolean?

rmn30: When I did the mips version of this change I kept it for convenience (quite a few places refer to cap.sealed) and kept it synchronised with type field, although it is probably more appropriate to replace it with an isSealed function.

PR #18, opened 2020-05-15
Introduce capability load generation traps

nwf:
First stab at adjusting PTW for capability load generation bits.


src/cheri_pte.sail
```diff
@@ -15,12 +15,29 @@ bitfield PTE_Bits : pteAttribs = {
 
 /* Reserved PTE bits could be used by extensions on RV64.  There are
  * no such available bits on RV32, so these bits will be zeros on RV32.
+ *
+ * XXX Because these bits are 0 on RV32, do we want to use inverted polarity,
+ * so that 0 is the *enabled* form?
```
jrtc27, line 20: That's what we also do on MIPS, and for both these reasons I was earlier saying the same thing to Peter.

nwf:
For FETT, I think we want to just have `SC` and `LC`, taking the other two (`LC_Mod` and `LC_CLG`) to be constant `0`s.

Speaking of those bits, I'm not elated with the current names... I thought about `LC_CLG` and `LC_LCLG` but those aren't great either.  Open to suggestions.

src/cheri_sys_regs.sail
```diff
@@ -5,6 +5,8 @@
 bitfield ccsr : xlenbits = {
   cap_idx : 15 .. 10,
   cause   :  9 .. 5, /* cap cause */
+  gclgu   :  3,
+  gclgs   :  2,
   d       :  1,      /* dirty  */
```
pmundkur, line 10: comments here explaining these bits would be good.

src/cheri_pte.sail
```diff
@@ -15,12 +15,29 @@ bitfield PTE_Bits : pteAttribs = {
...
+ *
  */
 type extPte = bits(10)
 
 bitfield Ext_PTE_Bits : extPte = {
   LoadCap  : 9,
   StoreCap : 8,
+  LoadCap_Mod : 7,
+  LoadCap_CLG : 6,
 }
```
pmundkur, line 41: some explanatory comments here too would be good.

nwf:
I have added a little more commentary and shuffled the different behaviours so that `LC_Mod` is more obviously a *modification* to the `LC` behaviour and `LC_LCLG` is now just the `LCLG` until we have other need for it.

At the moment the code throws if an invalid state is found, but perhaps there's a more correct in-model thing to do.  What happens if the RWX bits are set to W or WX, for example?

nwf:
This batch of commits moves us to LoadCapInhibit and StoreCapInhibit bits and shuffles things a bit more so that, again, all-zeros should be what we want.  The Load side is more fully fleshed out than the Store side, which, right now, defines but does not use a "fast" capability-dirtying path.

nwf:
This is atop #19 and #20 and should, I think, get both the load and store sides.

nwf:
Orthogonality of commits is hard; where'd those pullbacks of patches go, anyway?  This PR now also includes fixes to `cheri_insts` to pass virtual, not physical, address to `handle_mem_exception` as well.

src/cheri_pte.sail
```diff
@@ -13,14 +13,87 @@ bitfield PTE_Bits : pteAttribs = {
...
+ * not enabled) and SC=1 SC_Mod=1 (stores succeed).
  */
 type extPte = bits(10)
 
 bitfield Ext_PTE_Bits : extPte = {
-  LoadCap  : 9,
-  StoreCap : 8,
+  StoreCap     : 9, /* Permit capability stores */
+  LoadCap      : 8, /* Permit capability loads */
+  StoreCap_Mod : 7, /* Modify capability store inhibit; see above table */
```
bsdjhb, line 94: The use of inhibit here and in the line below is perhaps stale now?

src/cheri_types.sail
```diff
@@ -94,7 +93,7 @@ function CapExCode(ex) : CapEx -> bits(5) =
     CapEx_ReturnTrap                    => 0b00110,
     CapEx_TSSUnderFlow                  => 0b00111,
     CapEx_UserDefViolation              => 0b01000,
-    CapEx_TLBNoStoreCap                 => 0b01001,
```
bsdjhb, line 97: At some point we will want to annotate this CHERI exception code as MIPS-only in the ISA doc.

src/cheri_pte.sail
nwf: Yup, good eye.

src/cheri_pte.sail
```diff
@@ -39,7 +85,62 @@ function isInvalidPTE(p : pteAttribs, ext : extPte) -> bool = {
...
+
+/*
+ * Assuming we're allowed to load from this page, modulate our cap response
+ */
+val checkPTEPermission_LC : (PTE_Bits, Ext_PTE_Bits, ext_ptw) -> PTE_Check effect { escape, rreg }
+function checkPTEPermission_LC(p, e, ext_ptw) = {
+  let lcav : ext_ptw_lc =
+    if e.CapRead() == 0b0
+    then match (e.CapRead_Gen(), e.CapRead_Mod()) {
+           (0b1, _)  => not_implemented("Reserved PTE extension bits: !LC but Gen"),
```
PeterRugg, line 99: Does not_implemented imply a simulator crash? Surely we want to define some behaviour here, e.g. trap always?

nwf:
Here's a slightly cleaned up PR that also addresses @PeterRugg's observation that `not_implemented` is not terribly sociable.

src/cheri_pte.sail
```diff
@@ -49,6 +88,62 @@ union PTE_Check = {
...
+         then { if e.CapRead_Gen() == 0b0
+                then PTE_Check_Success(ext_ptw_lc_join(ext_ptw, PTW_LC_OK)) /* Unmodified LC case: go ahead */
+                else PTE_Check_Failure(ext_ptw, EPTWF_CAP_ERR)
+         }
+         else { /* With CapRead_Mod on with LC, use Gen */
+                let gclg : bits(1) = if p.U() == 0b1 then sccsr.gclgu() else sccsr.gclgs();
+                if e.CapRead_Gen() == gclg
+                then PTE_Check_Success(ext_ptw_lc_join(ext_ptw, PTW_LC_OK))
+                else PTE_Check_Success(ext_ptw_lc_join(ext_ptw, PTW_LC_TRAP))
+         };
```
jrtc27, line 113: This would be better as a big 3-tuple match with the fields and cases in the same order as your helpful table above. As it stands it's hard to match things up, and a bit messy.

src/cheri_pte.sail
```diff
@@ -65,6 +88,26 @@ union PTE_Check = {
...
+ */
+val checkPTEPermission_LC : (PTE_Bits, Ext_PTE_Bits, ext_ptw) -> PTE_Check effect { escape, rreg }
+function checkPTEPermission_LC(p, e, ext_ptw) =
+  match (e.CapRead(), e.CapRead_Mod(), e.CapRead_Gen()) {
+    (0b0, 0b0, 0b0)  => PTE_Check_Success(ext_ptw_lc_join(ext_ptw, PTW_LC_CLEAR)), /* Clear tag for "unmodified" no-LC case */
+    (0b0, 0b1, 0b0)  => PTE_Check_Success(ext_ptw_lc_join(ext_ptw, PTW_LC_TRAP)),  /* Trap on tag load for "modified" no-LC case */
+    (0b0, _  , 0b1)  => PTE_Check_Failure(ext_ptw, EPTWF_CAP_ERR),
+    (0b1, 0b0, 0b0)  => PTE_Check_Success(ext_ptw_lc_join(ext_ptw, PTW_LC_OK)), /* Unmodified LC case: go ahead */
+    (0b1, 0b0, 0b1)  => PTE_Check_Failure(ext_ptw, EPTWF_CAP_ERR),
+    (0b0, 0b1, lclg) => {
```
PeterRugg, line 102: Should be (1, 0, lclg)?

nwf: Whoop, yes, how embarrassing.

src/cheri_sys_regs.sail
```diff
@@ -5,6 +5,8 @@
 bitfield ccsr : xlenbits = {
   cap_idx : 15 .. 10,
   cause   :  9 .. 5, /* cap cause */
+  gclgu   :  3,      /* Global Cap Load Gen bit for User pages */
+  gclgs   :  2,      /* Global Cap Load Gen bit for System (not User) pages */
```
PeterRugg, line 9: It would seem this adds the bits per privilege level, rather than only for sccsr

PeterRugg: Also, it seems I can't comment on lines that weren't actually changed, but the legalize_ccsr function probably needs updating to allow changes to these bits.
I'm also surprised that sail still does exception cause/reg in CCSR (@jrtc27 should I go through and make that change?)

src/cheri_pte.sail
```diff
@@ -49,6 +88,57 @@ union PTE_Check = {
...
 
+/*
+ * Assuming we're allowed to load from this page, modulate our cap response
+ */
+val checkPTEPermission_LC : (PTE_Bits, Ext_PTE_Bits, ext_ptw) -> PTE_Check effect { escape, rreg }
+function checkPTEPermission_LC(p, e, ext_ptw) = {
+  let res : PTE_Check = match (e.CapRead(), e.CapRead_Mod(), e.CapRead_Gen()) {
+    (0b0, 0b0, 0b0)  => PTE_Check_Success(ext_ptw_lc_join(ext_ptw, PTW_LC_CLEAR)), /* Clear tag for "unmodified" no-LC case */
+    (0b0, 0b1, 0b0)  => PTE_Check_Success(ext_ptw_lc_join(ext_ptw, PTW_LC_TRAP)),  /* Trap on tag load for "modified" no-LC case */
+    (0b0, _  , 0b1)  => PTE_Check_Failure(ext_ptw, EPTWF_CAP_ERR),
```
jrtc27, line 99:
```suggestion
    (0b0, _  , 0b1)  => PTE_Check_Failure(ext_ptw, EPTWF_INVALID),
```
otherwise this will hit:
https://github.com/CTSRD-CHERI/sail-cheri-riscv/blob/7cfc2d1b148f020e0ab51ddfa8bd7f9a5f23e22f/src/cheri_ptw.sail#L44
(though maybe we want to distinguish these as capability page faults still?)

src/cheri_pte.sail
```diff
@@ -49,6 +88,57 @@ union PTE_Check = {
...
+    (0b1, 0b0, 0b0)  => PTE_Check_Success(ext_ptw_lc_join(ext_ptw, PTW_LC_OK)), /* Unmodified LC case: go ahead */
+    (0b1, 0b0, 0b1)  => PTE_Check_Failure(ext_ptw, EPTWF_CAP_ERR),
```
jrtc27, line 101:
```suggestion
    (0b1, 0b0, 0b1)  => PTE_Check_Failure(ext_ptw, EPTWF_INVALID),
```
Ditto.

jrtc27 (changes requested):

src/cheri_pte.sail
```diff
@@ -77,29 +167,57 @@ function checkPTEPermission(ac : AccessType(ext_access_type), priv : Privilege,
...
+  let res : PTE_Check =
+    match res {
+      PTE_Check_Failure(_, _) => res,
+      PTE_Check_Success(_) =>
+       if scisa.pte_cap_crmod() == 0b0 &
+          not(e.CapRead_Mod() == 0b0 &
+              e.CapRead_Gen() == 0b0)
+       then PTE_Check_Failure(ext_ptw, EPTWF_INVALID)
+       else res
+    };
```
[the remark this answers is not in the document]
nwf: No, but it plausibly could be.  I'd meant it to be a more general check for bad settings of capability bits given `scisa` but it did end up just being load-side behaviours.

src/cheri_pte.sail
```diff
@@ -105,8 +105,39 @@ function checkPTEPermission_LC(p, e, ext_ptw) =
...
+}
 
+val checkPTEPermission_SC : (Ext_PTE_Bits, ext_ptw) -> PTE_Check effect { rreg }
+function checkPTEPermission_SC(e, ext_ptw) = {
+  if   e.CapWrite() == 0b1
+  then PTE_Check_Success(ext_ptw)
+  else {
+         let ext_ptw' : ext_ptw = ext_ptw_sc_join(ext_ptw, PTW_SC_TRAP);
+
+         if   scisa.pte_cap_sc_datadep() == 0b1
```
PeterRugg, line 136: Should this line check the tag?

PeterRugg: Oh, sorry, I see that we return a value indicating that a tag should imply a trap, but I don't see where the check of that tag actually happens.

src/cheri_pte.sail
nwf: Will we?  I thought we'd hit the `EPTWF_CAP_ERR` case of  `ext_get_ptw_error` and then the `AT_CAP_ERR` cases of `translationException`?

src/cheri_pte.sail
nwf: Indeed, I think this doesn't work as I placed it here.  The data dependence is supposed to be brought forward into translation by the access type (`Write(Data)` vs `Write(Cap)`) and is ultimately analysed by `translationException` and so the right place for `pte_cap_sc_datadep` is in `cheri_instr`.  One revised commit coming right up.

src/cheri_pte.sail
jrtc27: Not for `Read`, only `Write` and `ReadWrite` (Peter and I observed this with his in-progress TestRIG translation template).

nwf:
Thanks for feedback.  Let's try this new series on for size?

src/cheri_pte.sail
```diff
@@ -16,23 +16,62 @@ bitfield PTE_Bits : pteAttribs = {
...
- * from or stores to pages in the SV32 MMU).
+ * For capability store the behaviors are as follows.
+ *
+ *   CW  CD  Action
+ *
+ *    0   X    Trap on tagged capability store
+ *    1   0    CAS the PTE to CW=1, CD=1 (i.e., set both bits) or trap
+ *    1   1    Permit tagged capability store
+ *
+ * The CW/CD pair is akin to SVnn's W/D bits.
```
jrtc27, line 27: The official names are Sv32 etc; the capital V only comes from C macro naming conventions.

src/cheri_insts.sail
```diff
@@ -1346,10 +1346,27 @@ function clause execute (LoadDataCap(rd, cs1, is_unsigned, width)) = {
   handle_load_data_via_cap(rd, 0b0 @ cs1, cs1_val, vaddr, is_unsigned, width)
 }
 
+union load_cap_ptw_ext_result = {
+  LCPE_OK : Capability,
+  LCPE_Exn : ExceptionType
+}
+
+val handle_load_cap_ptw_ext : (Capability, bool, ext_ptw_lc) -> load_cap_ptw_ext_result
+function handle_load_cap_ptw_ext (v, vialc, ptw_info) =
```
jrtc27, line 1355: I'd prefer this function be inlined; it's only used twice (maybe 3 times with AMOs) so the code size is likely going to be no different, and it avoids having more magic helpers. I feel currently this just makes the code harder to read, especially with the Permit_Load_Cap handling being pushed down into here.

jrtc27 (changes requested):

src/cheri_sys_regs.sail
```diff
@@ -23,6 +25,8 @@ function legalize_ccsr(c : ccsr, v : xlenbits) -> ccsr = {
   let v = Mk_ccsr(v);
   let c = update_cap_idx(c, v.cap_idx());
   let c = update_cause(c, v.cause());
+  let c = update_gclgu(c, v.gclgu());
+  let c = update_gclgs(c, v.gclgs());
```
jrtc27, line 29: Either the bits need to not exist for non-S-mode or we need a single register with per-mode views a la mstatus such that the bits are only accessible for >= S-mode. I'm a fan of the latter given it mirrors how a lot of the CSRs work; they tend to only have distinct registers when you can't cram all the per-mode bits into a single one. In that case the bit names should also have an `s` prefix to indicate the least-privileged mode that can see them, and I'd personally be inclined to move the `s` and `u` suffixes towards the front too to give `sugclg` and `ssgclg`, it feels more RISC-V-y to me, though there isn't really an equivalent to base it on.

src/cheri_pte.sail
```diff
@@ -77,29 +135,42 @@ function checkPTEPermission(ac : AccessType(ext_access_type), priv : Privilege,
     (_,            Machine)    => internal_error("m-mode mem perm check")
   };
 
+  let res : PTE_Check =
```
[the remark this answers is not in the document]
nwf: Sail seems to disagree?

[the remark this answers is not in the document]
jrtc27: Hmph, what about the ones later? Is it a quirk of enums?

src/cheri_sys_regs.sail
nwf: OK, but since we think that `xccsr.cap_cause` and `.cap_idx` are going away, can that happen before I refactor all this?

src/cheri_sys_regs.sail
```diff
@@ -32,11 +43,11 @@ function legalize_ccsr(c : ccsr, v : xlenbits) -> ccsr = {
...
 function clause ext_read_CSR (0x9C0) = Some(sccsr.bits())
-function clause ext_read_CSR (0xBC0) = Some(mccsr.bits())
+function clause ext_read_CSR (0xBC0) = Some(sccsr.bits())
 
-function clause ext_write_CSR (0x8C0, value) = { uccsr = legalize_ccsr(uccsr, value); Some(uccsr.bits()) }
-function clause ext_write_CSR (0x9C0, value) = { sccsr = legalize_ccsr(sccsr, value); Some(sccsr.bits()) }
-function clause ext_write_CSR (0xBC0, value) = { mccsr = legalize_ccsr(mccsr, value); Some(mccsr.bits()) }
+function clause ext_write_CSR (0x8C0, value) = { uccsr = legalize_uccsr(uccsr, value); Some(uccsr.bits()) }
+function clause ext_write_CSR (0x9C0, value) = { sccsr = legalize_sccsr(sccsr, value); Some(sccsr.bits()) }
+function clause ext_write_CSR (0xBC0, value) = { sccsr = legalize_sccsr(sccsr, value); Some(sccsr.bits()) }
```
jrtc27, line 50: Maybe nicer to pass in Machine/Supervisor/User rather than duplicating the function?

src/cheri_sys_regs.sail
```diff
@@ -3,22 +3,33 @@
...
-register mccsr : ccsr
-register sccsr : ccsr
+register sccsr : ccsr /* mccsr aliases sccsr */
 register uccsr : ccsr
 
 /* access to CCSRs */
 
-// for now, use a single privilege-independent legalizer
-function legalize_ccsr(c : ccsr, v : xlenbits) -> ccsr = {
+function legalize_ccsr(p : Privilege, c : ccsr, v : xlenbits) -> ccsr = {
```
jrtc27, line 17: Maybe `cp`? Otherwise it could be confused for what privilege the hart is currently running with.

nwf: I went with the slightly longer `csrp` since `cp` might still be read as `current privilege` at a glance.

nwf:
Rebased.  These continue to pass https://gist.github.com/nwf/cd4d760247ee54e3c9e136ba53d9fdbf ("pte-bits.S")

nwf:
As per https://github.com/CTSRD-CHERI/qemu/pull/150#discussion_r653538537, @jrtc27 makes an argument that the behaviour of `CLoadTags` should rather be that it doesn't trap but instead clears tags if the PTW says `PTW_LC_CLEAR`.  Data dependence for the `PTW_LC_TRAP` case seems perhaps a bridge too far, but could be warranted as well.

src/cheri_pte.sail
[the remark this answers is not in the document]
nwf: (Very) Belatedly, I'm not sure what it's a quirk of, beyond `sail`, but the type decls when assigning from `match` clauses seem required; the first `let res`, which assigns from an `if` expression does not seem to need a type decl and ti's gone in the latest patch.

src/cheri_insts.sail
```diff
@@ -1495,11 +1495,25 @@ function handle_load_cap_via_cap(cd, auth_idx, auth_val, vaddrBits) = {
...
+          match ptw_info.ptw_lc {
+            PTW_LC_OK => {
+              C(cd) = { v with tag = v.tag & auth_val.permit_load_cap };
+              RETIRE_SUCCESS
+            },
+            PTW_LC_CLEAR => {
+              C(cd) = { v with tag = false };
+              RETIRE_SUCCESS
+            },
+            PTW_LC_TRAP  => {
```
jrtc27, line 1507: PTW_LC_OK isn't made to line up but this one is (repeated for the various copies of this)

src/cheri_addr_checks.sail
```diff
@@ -212,6 +212,12 @@ function ext_handle_data_check_error(err : ext_data_addr_error) -> unit = {
   handle_cheri_cap_exception(capEx, regnum)
 }
 
+/* Default implementations of these hooks permit all accesses.  */
+function ext_check_phys_mem_read (access_type, paddr, size, aquire, release, reserved, read_meta) =
+  Ext_PhysAddr_OK ()
```
jrtc27, line 217: Drop the space (again below)

src/cheri_pte.sail
```diff
@@ -78,23 +78,56 @@ bitfield PTE_Bits : pteAttribs = {
...
+ * The CW/CD pair is akin to Sv{32,48,...}'s W/D bits.
+ *
+ * The CW=1 CD=0 behavior is described as a CAS (and not an AMO OR) of the PTE
+ * to close race conditions, much as with W and D.  Should the TLB have this
+ * state cached and then observe a CW=1 CD=1 PTE, no PTE write is necessary.
+ * On the other hand, if CW=0 is observed, the store operation must trap.
+ *
+ * The intended semantics for capability loads are as follows.
+ *
+ *   CR  CR_Trap CR_Gen Action
```
jrtc27, line 98: Bit weird to abbreviate CR but not Trap and Gen. The mixing of camel case and underscores is a bit odd, too.

src/cheri_pte.sail
```diff
@@ -78,23 +78,56 @@ bitfield PTE_Bits : pteAttribs = {
...
+ * Sv32: There are no extension bits available, so we hard-code the result to
+ * CW=1 CR=1 CD=1 CR_Trap=0 CR_Gen=0
  */
 type extPte = bits(10)
 
 bitfield Ext_PTE_Bits : extPte = {
   CapWrite     : 9, /* Permit capability stores */
   CapRead      : 8, /* Permit capability loads */
+  CapDirty     : 7, /* Capability Dirty flag */
+  CapRead_Trap : 6, /* Trap on capability loads; see above table */
```
jrtc27, line 118: Drop the underscores?

src/cheri_pte.sail
```diff
@@ -103,14 +136,44 @@ function isPTEPtr(p : pteAttribs, ext : extPte) -> bool = {
...
+  (e.CapRead() == 0b1 & e.CapRead_Gen() == 0b1 & e.CapRead_Trap() == 0b0)
 }
 
 union PTE_Check = {
   PTE_Check_Success : ext_ptw,
   PTE_Check_Failure : (ext_ptw, ext_ptw_fail)
 }
 
+/* Gate Cap-transporting stores on the CapWrite PTE bit */
+val checkPTEPermission_SC : (Ext_PTE_Bits, ext_ptw) -> PTE_Check effect { rreg }
```
jrtc27, line 151: I don't think it does have any effects?

src/cheri_pte.sail
```diff
@@ -176,9 +252,21 @@ function update_PTE_Bits(p : PTE_Bits, a : AccessType(ext_access_type), ext : ex
...
-  if update_d | update_a then {
+  let pte_ext = Mk_Ext_PTE_Bits(ext);
+
+  // store cap bits
+  let update_cd = pte_ext.CapWrite() == 0b1 &
+                  pte_ext.CapDirty() == 0b0 &
+                  (match a {
+                    Write(Cap)        => true,
+                    ReadWrite(_, Cap) => true,
+                    _                 => false
```
jrtc27, line 263: For checkPTEPermission you exhaustively enumerate the options, but not here?

src/cheri_sys_regs.sail
```diff
@@ -65,22 +65,33 @@
...
+  ugclg   :  3,      /* Global Cap Load Gen bit for User pages */
+  sgclg   :  2,      /* Global Cap Load Gen bit for System (not User) pages */
   d       :  1,      /* dirty  */
   e       :  0       /* enable */
 }
 
-register mccsr : ccsr
-register sccsr : ccsr
+register sccsr : ccsr /* mccsr aliases sccsr */
 register uccsr : ccsr
```
jrtc27, line 75: Bit weird to still have uccsr... I kinda get why, but that's primarily because uccsr currently serves no purpose. Also, if mccsr and sccsr alias then normally you'd have the definition be mccsr and sccsr be an alias for (or view into) it, not the other way round.

nwf: Well, it "serves no purpose" because the `d` and `e` bits are not meaningfully implemented in sail, but they're still defined...

nwf: Though maybe that does argue against making `sccsr` and `mccsr` alias.

nwf: I'm not particularly happy with what I have just written in `cheri_pte.sail` -- "use `sccsr` if `haveSupMode()` and `mccsr` otherwise" -- but it might be viable.  I'm quite open to suggestions.

nwf: @jrtc27 Opinions?

src/cheri_pte.sail
```diff
@@ -78,23 +78,57 @@ bitfield PTE_Bits : pteAttribs = {
...
+ *                   the compared bit is keyed off of this PTE's U.  sccsr is
+ *                   used if the machine has Supervisor mode, otherwise mccsr.
+ *
+ * Sv32: There are no extension bits available, so we hard-code the result to
+ * CW=1 CR=1 CD=1 CRT=0 CRG=0
  */
 type extPte = bits(10)
 
 bitfield Ext_PTE_Bits : extPte = {
   CapWrite     : 9, /* Permit capability stores */
```
arichardson, line 116: This is pre-existing, but bits 9 and 8 conflict with “Svpbmt” Standard Extension for Page-Based Memory Types. I wonder if we should shift everything down by two as part of this change series or start from zero (since upstream started from the top)?

nwf: Ouch.  I guess that's not the worst possible flag day, it'd just touch sail, the cores, and the CheriBSD kernel.

PR #19, opened 2020-05-29
Refactor PTE extension logic

nwf:
Notably, ext_ptw is now a struct (a pair, at the moment) rather than a
single enumerated value.  This should make it easier to do the
CLG/capdirty tracking bits.

No functional change intended.

### DIFF
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -909,7 +909,7 @@ function handle_load_cap_via_cap(rd, cs, cap_val, vaddrBits) = {
       let c = mem_read_cap(addr, aq, rl, false);
       match c {
         MemValue(v) => {
-          let cr = if ptw_info == PTW_LOAD_CAP_ERR
+          let cr = if   ptw_info.ptw_lc_mod == PTW_LCM_CLR
                    then {v with tag = false} /* strip the tag */
                    else {v with tag = v.tag & cap_val.permit_load_cap};
           writeCapReg(rd, cr);
@@ -1007,7 +1007,7 @@ if not(cap_val.tag) then {
       let c = mem_read_cap(addr, aq, rl, false);
       match c {
         MemValue(v) => {
-          let cr = if   ptw_info == PTW_LOAD_CAP_ERR
+          let cr = if   ptw_info.ptw_lc_mod == PTW_LCM_CLR
                    then {v with tag = false} /* strip the tag */
                    else {
                      /* the Sail model currently reserves virtual addresses */
--- a/src/cheri_pte.sail
+++ b/src/cheri_pte.sail
@@ -71,37 +71,24 @@ function checkPTEPermission(ac : AccessType(ext_access_type), priv : Privilege, 
   };
 
   let e = Mk_Ext_PTE_Bits(ext);
-  let (succ, pte_chk) : (bool, pte_check) =
+  let lcav = if e.LoadCap()  == 0b1 then PTW_LCM_OK else PTW_LCM_CLR;
+  let scav = if e.StoreCap() == 0b1 then PTW_SCM_OK else PTW_SCM_TRAP;
+  let (succ, ext_ptw') : (bool, ext_ptw) =
   match (base_succ, ac) {
     /* Base translation exceptions take priority over CHERI exceptions */
-    (false, _)               => (false, PTE_CAP_OK),
+    (false, _)               => (false, init_ext_ptw),
 
-    (true,  Read(Cap))       => if   e.LoadCap() == 0b1
-                                then (true, PTE_CAP_OK)
-                                /* Allow the address translation to proceed, but mark for tag stripping */
-                                else (true, PTE_LOAD_CAP_ERR),
+    (true,  Read(Cap))       => (true, ext_ptw_join_lcm(ext_ptw, lcav)),
+    (true,  Write(Cap))      => (true, ext_ptw_join_scm(ext_ptw, scav)),
+    (true,  ReadWrite(Cap))  => (true, ext_ptw_join_scm(ext_ptw_join_lcm(ext_ptw, lcav), scav)),
 
-    (true,  Write(Cap))      => if   e.StoreCap() == 0b1
-                                then (true, PTE_CAP_OK)
-                                /* Do not allow the address translation to proceed */
-                                else (false, PTE_STORE_CAP_ERR),
-
-    (true,  ReadWrite(Cap))  => if   e.StoreCap() == 0b1 & e.LoadCap() == 0b1
-                                then (true, PTE_CAP_OK)
-                                else if e.StoreCap() == 0b0
-                                /* Do not allow the address translation to proceed */
-                                then (false, PTE_STORE_CAP_ERR)
-                                /* Allow the address translation to proceed, but mark for tag stripping */
-                                else (true, PTE_LOAD_CAP_ERR),
-
-    (true,  Read(Data))      => (true, PTE_CAP_OK),
-    (true,  Write(Data))     => (true, PTE_CAP_OK),
-    (true,  ReadWrite(Data)) => (true, PTE_CAP_OK),
-    (true,  Execute())       => (true, PTE_CAP_OK)
+    (true,  Read(Data))      => (true, ext_ptw),
+    (true,  Write(Data))     => (true, ext_ptw),
+    (true,  ReadWrite(Data)) => (true, ext_ptw),
+    (true,  Execute())       => (true, ext_ptw)
   };
 
-  let ext_ptw = ext_accum_ptw_result(ext_ptw, pte_chk);
-  if succ then PTE_Check_Success(ext_ptw) else PTE_Check_Failure(ext_ptw)
+  if succ then PTE_Check_Success(ext_ptw') else PTE_Check_Failure(ext_ptw')
 }
 
 function update_PTE_Bits(p : PTE_Bits, a : AccessType(ext_access_type), ext : extPte) -> option((PTE_Bits, extPte)) = {
--- a/src/cheri_ptw.sail
+++ b/src/cheri_ptw.sail
@@ -24,21 +24,43 @@ function ptw_error_to_str(e) =
 
 overload to_str = {ptw_error_to_str}
 
+/*
+ * Produce the PTW_Error for PTE_Check_Failure-s arising from
+ * checkPTEPermission calls.
+ *
+ * We analyse only the store-side component of ext_ptw here.  The MMU/PTW logic
+ * does not see into the data path, and so load-side responses, including
+ * traps, must be handled by the instruction that initiated the translation by
+ * considering the ext_ptw result provided by the MMU/PTW.
+ */
 function ext_get_ptw_error(ext_ptw : ext_ptw) -> PTW_Error =
-  match (ext_ptw) {
-    PTW_CAP_OK        => PTW_No_Permission(),
-    PTW_LOAD_CAP_ERR  => PTW_Ext_Error(AT_CAP_ERR),
-    PTW_STORE_CAP_ERR => PTW_Ext_Error(AT_CAP_ERR)
+  match (ext_ptw.ptw_sc_mod) {
+    PTW_SCM_TRAP => PTW_Ext_Error(AT_CAP_ERR),
+    PTW_SCM_OK   => PTW_No_Permission()
   }
 
 /* conversion of these translation/PTW failures into architectural exceptions */
 function translationException(a : AccessType(ext_access_type), f : PTW_Error) -> ExceptionType = {
   let e : ExceptionType =
   match (a, f) {
-    /* First handle the vmem extension errors */
-    (ReadWrite(Cap), PTW_Ext_Error(AT_CAP_ERR)) => E_Extension(EXC_SAMO_CAP_PAGE_FAULT),
-    (Read(Cap), PTW_Ext_Error(AT_CAP_ERR))      => E_Extension(EXC_LOAD_CAP_PAGE_FAULT),
+
+    /* Tag-asserting stores can raise CHERI page faults. */
     (Write(Cap), PTW_Ext_Error(AT_CAP_ERR))     => E_Extension(EXC_SAMO_CAP_PAGE_FAULT),
+
+    /*
+     * XXX ReadWrite(Cap) does not provide enough information to distinguish
+     * between a tag-clearing store and tag-asserting store in the presence of
+     * a tag-capable load, or a tag-asserting store in the case of a
+     * tag-incapable load.  (That is, it only clearly distinguishes a
+     * tag-incapable load with a tag-clearing store from every other option.)
+     * Groundwork for fixing this exists in the sail RISC-V model, but we have
+     * not yet caught up to that point.  Fortunately, we don't actually have
+     * amoswap.c yet, so this is not observably wrong.
+     */
+    (ReadWrite(Cap), PTW_Ext_Error(AT_CAP_ERR)) => E_Extension(EXC_SAMO_CAP_PAGE_FAULT),
+
+    /* No other operations should raise CHERI-specific page faults */
+    (_, PTW_Ext_Error(_)) => internal_error("Unexpected PTW Extension Error"),
 
     /* For other exceptions, Cap accesses fault similar to Data accesses below. */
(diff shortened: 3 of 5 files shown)


src/cheri_riscv_types.sail
```diff
@@ -1,28 +1,53 @@
...
+ * We are going to carry around a summary of the PTW walk for data-dependent
+ * (well, tag-dependent) behaviors by the instructions.
+ *
+ */
 
-/* Extension for CHERI page-table-walks */
-
-enum ext_ptw = {PTW_CAP_OK, PTW_LOAD_CAP_ERR, PTW_STORE_CAP_ERR}
+enum ext_ptw_lcm = {
+  PTW_LCM_OK /* Tags flow as might be expected */,
```
jrtc27, line 10: I'd put commas before the comment, on first read it looked like they were missing.

src/cheri_pte.sail
```diff
@@ -38,102 +38,99 @@ union PTE_Check = {
...
                                        }
-                                       else (false, PTE_CAP_OK)
+                                       else (false, ext_ptw)
                                      },
 
     (ReadWrite(Cap),  User)       => { if   p.U() == 0b1 & p.R() == 0b1 & p.W() == 0b1
                                        then {
                                          let  e = Mk_Ext_PTE_Bits(ext);
                                          if   e.StoreCap() == 0b1 & e.LoadCap() == 0b1
-                                         then (true, PTE_CAP_OK)
```
jrtc27, line 88: Should this whole thing not be `ext_ptw_join_scm(ext_ptw_join_lcm(ext_ptw, e.LoadCap() == 0b1 ? PTW_LCM_OK, PTW_LCM_CLR), e.StoreCap() == 0b1 ? PTW_SCM_OK, PTW_SCM_TRAP)`, with suitable local variables of course to make that more readable? This is currently a bug I believe:

```
cmove ca1, cnull
amoswap.c ca1, (a0)
```

Currently, that will allow you to bypass the tag clearing loads, as we'll just get back the `PTW_SCM_TRAP` and then not trap because the value being stored isn't tagged.

For the Read/Write cases I'd do similar things and push the conditional inside the argument of `ext_ptw_join_[ls]cm`, rather than conditionally call it. I think that's neater.

nwf: Yes, that's probably right.  Standby.

nwf: Mm, `amoswap.c` doesn't seem to be implemented, and I don't know that it would have hit that case even so, since it should still have issued a `ReadWrite(Cap)` and `PTW_SCM_TRAP` is always paired with a failing translation, but I think it's cleaner your way anyway. :)

src/cheri_pte.sail
```diff
@@ -38,102 +38,107 @@ union PTE_Check = {
...
-                                         else if e.StoreCap() == 0b0
-                                         /* return a failure since we should cause an exception */
-                                         then (false, PTE_STORE_CAP_ERR)
-                                         /* return a success since the translation should proceed */
-                                         else (true, PTE_LOAD_CAP_ERR)
+                                         let    e = Mk_Ext_PTE_Bits(ext);
+                                         let oksc = e.StoreCap() == 0b1;
+                                         let oklc = e.LoadCap() == 0b1;
+                                         ( /* Translation succeeds if SC also OK */
+                                           oksc,
```
jrtc27, line 90: Isn't this wrong? Translation succeeds, you just need to do a data-dependent trap. Similarly with the plain Write and Supervisor cases.

src/cheri_riscv_types.sail
```diff
@@ -1,28 +1,57 @@
...
+ *
+ * We are going to carry around a summary of the PTW walk for data-dependent
+ * (well, tag-dependent) behaviors by the instructions.
+ *
+ */
 
-/* Extension for CHERI page-table-walks */
+enum ext_ptw_lcm = {
+   /* Tags flow as might be expected */
+  PTW_LCM_OK,
```
jrtc27, line 11: Indentation has gone weird for all these.

src/cheri_pte.sail
nwf: I was just following along the existing code for this patch, but yes, that's probably the intended behaviour.  (We could also get it by making the request type -- `Write(Data)` vs `Write(Cap)` -- depend on the tag, but since we're going to be expressing data-dependence as post-TLB for other things, consistency is probably a grand plan.)

nwf:
Sigh.  I seem to have definitely squashed together a bunch of commits due to weak git-fu.  Lemme straighten that out.

src/cheri_pte.sail
```diff
@@ -38,102 +38,69 @@ union PTE_Check = {
...
-                                           (true, PTE_LOAD_CAP_ERR)
-                                         }
-                                       }
-                                       else (false, PTE_CAP_OK)
+                                       then
+                                         /*
+                                          * Allow the address translation to proceed,
+                                          * but mark for tag stripping
+                                          */
+                                         (true, ext_ptw_join_lcm(ext_ptw, lcav))
```
jrtc27, line 64:
```suggestion
                                        /*
                                         * Allow the address translation to proceed,
                                         * but mark for tag stripping
                                         */
                                        then (true, ext_ptw_join_lcm(ext_ptw, lcav))
```
and other instances

nwf: I'd dropped these comments when I did some of my own tidying, it seems, so they're no longer present in the version just pushed.

src/cheri_insts.sail
```diff
@@ -1007,7 +1007,7 @@ if not(cap_val.tag) then {
       let c = mem_read_cap(addr, aq, rl, false);
       match c {
         MemValue(v) => {
-          let cr = if   ptw_info == PTW_LOAD_CAP_ERR
+          let cr = if   ptw_info.ptw_lc_mod == PTW_LCM_CLR
```
jrtc27, line 1010: It'd be nice if the type by this point wasn't such that, were there to be a PTW_LCM_TRAP, it'd show up. Really we just want a bool (or a struct containing a bool, since ptw_info is shared among all paths), so this just becomes `if ptw_info.lcm_clear`. Unfortunately since it's ext_ptw all the way through, `ext_get_ptw_error` needs enough info to know whether it was a CHERI page fault or not so that `translateException` knows which type of page fault to raise, which means that ptw_info has to have a bit of extra internal state, though that can be just another bool for whether to give `PTW_No_Permission` or `PTW_Ext_Error`.

src/cheri_insts.sail
```diff
@@ -1308,26 +1316,34 @@ function handle_store_cond_cap_via_cap(cs2, cs, cap_val, vaddrBits) = {
...
+                  MemValue(true)  => {
+                     writeCapReg(cs2, int_to_cap(EXTZ(0b0)));
+                     cancel_reservation();
+                     RETIRE_SUCCESS
+                  },
+                  MemValue(false) => {
+                     writeCapReg(cs2, int_to_cap(EXTZ(0b1)));
+                     cancel_reservation();
+                     RETIRE_SUCCESS
+                  },
```
jrtc27, line 1341: Surely whilst we're here we can make this code less crazy as:
```suggestion
                  MemValue(success)  => {
                     writeCapReg(cs2, int_to_cap(EXTZ(bool_to_bits(!success))));
                     cancel_reservation();
                     RETIRE_SUCCESS
                  },
```
and at some point maybe do the same to sail-riscv's A extension implementation (though `X(rd) = EXTZ(0b0);` is at least more succinct than `writeCapReg(cs2, int_to_cap(EXTZ(0b0)));`...).

src/cheri_insts.sail
nwf: Unfortunately, the `PTW_LCM_TRAP` *has* to show up here: it's tag-dependent and the PTW has no access to the data path.

src/cheri_insts.sail
```diff
@@ -1165,16 +1165,24 @@ function handle_store_cap_via_cap(rs, cs, cap_val, vaddrBits) = {
...
-        MemValue(_) => {
-          let res : MemoryOpResult(bool) = mem_write_cap(addr, rs_val, aq, rl, false);
-          match (res) {
-            MemValue(true)  => RETIRE_SUCCESS,
-            MemValue(false) => internal_error("store got false from mem_write_value"),
-            MemException(e) => { handle_mem_exception(addr, e); RETIRE_FAIL }
+    TR_Address(addr, caveats) => {
+      match (caveats.ptw_sc_mod, rs_val.tag) {
+        (PTW_SCM_TRAP, true) => {
+          handle_cheri_cap_exception(CapEx_TLBNoStoreCap, cs);
```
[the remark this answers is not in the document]
nwf: `PTW_SCM_TRAP` can be hidden in the PTW logic, but until ReadWrite has its load/store sides split, so that we can indicate whether the store is tag-asserting or tag-clearing (rather than merely being tag-capable), we will erroneously trap on amoswap.c that store a cleared tag.

src/cheri_insts.sail
nwf: That may be, but I think in light of your comment above, we are no longer here, so I'll leave this to a later cleanup.

nwf:
I have no idea how it was building, but it was when I tried... that's not encouraging.  In any case, here's a respin that should be more agreeable.

src/cheri_insts.sail
jrtc27: I didn't mean that the signal has to be here. I meant that it was here _in this form_ with potentially states we don't care about (ie, in the full LCM world, you have PTW_LCM_TRAP, PTW_LCM_CLR and PTW_LCM_OK, and all three are possible values according to the type here, but PTW_LCM_TRAP is known to never happen; really we just want a single bool by this point saying whether to clear or preserve the tag, as the trapping has already happened). But as I said above, that requires refactoring the sail-riscv ext_ptw code to split things out into different types at different stages, so this is fine for now.

src/cheri_ptw.sail
```diff
@@ -24,22 +24,58 @@ function ptw_error_to_str(e) =
...
+     * not yet caught up to that point.  Fortunately, we don't actually have
+     * amoswap.c yet, so this is not observably wrong.
+     */
+    (ReadWrite(Cap), PTW_Ext_Error(AT_CAP_ERR)) => E_Extension(EXC_SAMO_CAP_PAGE_FAULT),
+
+    /*
+     * XXX Read(_) and Execute() should never result in an ext_ptw value that
+     * gets mapped to PTW_Ext_Error(_); see checkPTEPermission.  Should we
+     * assert that here?
+     */
```
jrtc27, line 77: I think a `(_, PTW_Ext_Error(_)) => fatal error` wouldn't go amiss, yeah.

src/cheri_riscv_types.sail
```diff
@@ -1,28 +1,60 @@
...
-function ext_accum_ptw_result(ptw : ext_ptw, pte : pte_check) -> ext_ptw =
-  match (ptw, pte) {
-    (PTW_CAP_OK, PTE_CAP_OK)        => PTW_CAP_OK,
-    (PTW_CAP_OK, PTE_LOAD_CAP_ERR)  => PTW_LOAD_CAP_ERR,
-    (PTW_CAP_OK, PTE_STORE_CAP_ERR) => PTW_STORE_CAP_ERR,
+  /*
+   * Tag-asserting stores trap; tag-capable but -clearing instructions should
+   * indicate that they are performing Write(Data), not Write(Cap), operations
+   * to avoid spuriously raising CHERI page faults.
+   */
```
jrtc27, line 27: I don't think most of this documentation belongs here but on the definition of `Cap`.

src/cheri_ptw.sail
```diff
@@ -24,22 +24,58 @@ function ptw_error_to_str(e) =
...
+     * XXX ReadWrite(Cap) does not provide enough information to distinguish
+     * between a tag-clearing store and tag-asserting store in the presence of
+     * a tag-capable load, or a tag-asserting store in the case of a
+     * tag-incapable load.  (That is, it only clearly distinguishes a
+     * tag-incapable load with a tag-clearing store from every other option.)
+     * Groundwork for fixing this exists in the sail RISC-V model, but we have
+     * not yet caught up to that point.  Fortunately, we don't actually have
+     * amoswap.c yet, so this is not observably wrong.
+     */
+    (ReadWrite(Cap), PTW_Ext_Error(AT_CAP_ERR)) => E_Extension(EXC_SAMO_CAP_PAGE_FAULT),
```
jrtc27, line 71: Would you be willing to update us to the latest sail-riscv and fix this in a subsequent PR? :)

src/cheri_ptw.sail
```diff
@@ -24,22 +24,58 @@ function ptw_error_to_str(e) =
...
-    PTW_CAP_OK        => PTW_No_Permission(),
-    PTW_LOAD_CAP_ERR  => PTW_Ext_Error(AT_CAP_ERR),
-    PTW_STORE_CAP_ERR => PTW_Ext_Error(AT_CAP_ERR)
+  match (ext_ptw.ptw_sc_mod) {
+
+    /*
+     * This is confusing: PTW_No_Permission() is never used as such, but just
+     * dodges the pattern matches in translationException, below.
+     */
+    PTW_SCM_OK   => PTW_No_Permission(),
```
jrtc27, line 48: Well, yes and no. `PTW_No_Permission` is just the non-extension "page fault" error. Since `ext_get_ptw_error` has been called, we know by this point that translation has failed at a PTE, so if it's not a CHERI page fault then we know it must just be a normal one. It's not so much dodging pattern matches as being one of the many conditions that lead to a standard page fault (`PTW_Access` is the only case where a failing PTW can give a non-page fault trap); really, `translationException` is saying "{Fetch,Load,Store/AMO} page fault except for in a few special cases".

src/cheri_ptw.sail
```diff
@@ -24,22 +24,58 @@ function ptw_error_to_str(e) =
...
 
+/*
+ * I apologize for the confusion you have experienced if you have skipped over
+ * this comment and read the body of this function first.  Hopefully these two
+ * points resolve most of it:
+ *
+ * 1) While it may look like we're signaling that we must *always* raise
+ * faults, translationException below can see whether the instruction
+ * triggering the PTW is claiming to store an asserted tag, and will ignore our
+ * result if that isn't so.
```
jrtc27, line 35: We _are_ always raising faults? If we're not storing an asserted tag then we passed in `Data` so we'll `checkPTEPermission` will always say `PTW_SCM_OK`. And if we've been called then we know the PTE checking failed so we're definitely raising a (normal) page fault in that case.

nwf: I have taken another stab at saying what I meant to be saying.

src/cheri_ptw.sail
```diff
@@ -24,22 +24,62 @@ function ptw_error_to_str(e) =
...
 function ext_get_ptw_error(ext_ptw : ext_ptw) -> PTW_Error =
-  match (ext_ptw) {
-    PTW_CAP_OK        => PTW_No_Permission(),
-    PTW_LOAD_CAP_ERR  => PTW_Ext_Error(AT_CAP_ERR),
-    PTW_STORE_CAP_ERR => PTW_Ext_Error(AT_CAP_ERR)
+  match (ext_ptw.ptw_sc_mod) {
+    /*
+     * Raise an exception if this operation is Cap, not Data, as tested by
+     * translationException.
+     */
```
jrtc27, line 47: Is this _really_ necessary? It says PTW_SCM_TRAP, it's completely non-surprising that we're going to give a CHERI-related error.

src/cheri_ptw.sail
```diff
@@ -24,22 +24,62 @@ function ptw_error_to_str(e) =
...
+ * its way to translationException, below, so some conspiracy is possible: even
+ * though we cannot see the ext_access_type for the transaction at hand, and we
+ * did not carry it through from where it was visible to checkPTEPermission, it
+ * is again visible to translationException.
+ *
+ * We analyse only the store-side component of ext_ptw here.  The MMU/PTW logic
+ * does not see into the data path, and so load-side responses, including
+ * traps, must be handled by the instruction that initiated the translation by
+ * considering the ext_ptw result provided by the MMU/PTW.
+ */
```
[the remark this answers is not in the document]
nwf: Well, sure, not here, I guess, but it would have been incredibly helpful if there had been a kind of "information flow overview" of the PTW logic, as it's spread across multiple files and partially overridden between this and the baseline model.

nwf:
Obsoleted.

PR #20, opened 2020-06-05
PTW: All writes, not just of Data, dirty a page

nwf:


bauereiss:
Note that you might also want to double-check uses of `Data` in the base sail-riscv spec.  For example, the pattern match in `pmpCheckRWX` becomes incomplete when `ext_access_type` becomes an `enum` of more than `Data`.

pmundkur:
@bauereiss good catch.  it looks like `pmpCheckRWX` needs to provide a hook for extensions to customize. @nwf what would you like it to look like?

nwf:
> @bauereiss good catch. it looks like `pmpCheckRWX` needs to provide a hook for extensions to customize. @nwf what would you like it to look like?

I don't know much about the PMP.  Looking at section 3.6 of the privileged spec, tho', it looks like there's not a huge amount of room for extensibility... so I wonder if either...

1. Extensions should expect to replace `pmpCheckRWX` wholesale, and/or
2. (the default) `pmpCheckRWX` should match on the `AccessType` but not the `ext_access_type` (i.e., with `Execute()`, `Read(_)`, `Write(_)`, and `ReadWrite(_)`).

jrtc27:
> @pmundkur:
> 
> > @bauereiss good catch. it looks like `pmpCheckRWX` needs to provide a hook for extensions to customize. @nwf what would you like it to look like?
> 
> I don't know much about the PMP. Looking at section 3.6 of the privileged spec, tho', it looks like there's not a huge amount of room for extensibility... so I wonder if either...
> 
> 1. Extensions should expect to replace `pmpCheckRWX` wholesale, and/or
> 2. (the default) `pmpCheckRWX` should match on the `AccessType` but not the `ext_access_type` (i.e., with `Execute()`, `Read(_)`, `Write(_)`, and `ReadWrite(_)`).

I would assume 2, and that if we wanted to have PMP equivalents of the PTE bits we would add extra bits in CSRs somewhere. Though I would have said the same about the PTE code ideally matching `Read(_)` rather than `Read(Data)` by default...

pmundkur:
@jrtc27 @nwf ok, I've pushed option 2 (matching on `Read(_)` etc) for PMP to sail-riscv, since it is the simplest for now.

PR #21, opened 2020-06-07
First stab at sentries

nwf:


src/cheri_sys_exceptions.sail
```diff
@@ -47,6 +51,18 @@ function prepare_trap_vector(p : Privilege, c : Mcause) -> xlenbits = {
                            Supervisor => STCC,
                            User       => UTCC
                          };
+
+  /*
+   * XXX I'd prefer to use match here, but I don't understand how to shift
+   * types in the other direction, that is, to go from typeof(tcc.otype)
+   * to typeof(otype_unsealed).
+   */
+  let tcc = if      tcc.otype == to_bits(otype_width, otype_unsealed)
```
arichardson, line 60: I wonder if we should add helpers such as `capIsSentry`/`capIsUnsealed`/`capIsSealedWithType` (or similar) to cheri_cap_common.sail? Not sure about the naming conventions there, seems quite inconsistent.

src/cheri_insts.sail
```diff
@@ -748,6 +785,32 @@ function clause execute (CUnseal(cd, cs, ct)) =
...
+    RETIRE_FAIL
+  } else if (cs_val.sealed) then {
+    handle_cheri_reg_exception(CapEx_SealViolation, cs);
+    RETIRE_FAIL
+  } else if not (cs_val.permit_execute) then {
+    handle_cheri_reg_exception(CapEx_PermitExecuteViolation, cs);
+    RETIRE_FAIL
+  } else {
+    let (success, newCap) = sealCap(cs_val, to_bits(otype_width, otype_sentry));
+    if not (success) then {
```
arichardson, line 804: Since this can't happen should we use an assert instead?

src/cheri_sys_exceptions.sail
jrtc27: xTCC is fundamentally incompatible with sentries given you have a vectored mode. In theory you could support sentries for direct mode, as the low bits happen to be 0 so it's taken as-is, but this is the only possible exception and I don't particularly like that conditional behaviour.

src/cheri_insts.sail
```diff
@@ -826,9 +889,13 @@ function clause execute(CJALR(cd, cb)) =
   } else {
     let (success, linkCap) = setCapAddr(PCC, nextPC); /* Note that nextPC accounts for compressed instrucitons */
     assert(success, "Link cap should always be representable.");
-    writeCapReg(cd, linkCap);
+    assert(linkCap.otype == to_bits(otype_width, otype_unsealed),
+           "Link cap should always be unsealed");
```
jrtc27, line 893:
```suggestion
    assert(!linkCap.sealed, "Link cap should always be unsealed");
```

Since we have the summary bit.

jrtc27:
The difficulty is around xEPCC legalisation, which as far as I can tell you have not addressed at all (and I don't even see code to unseal xEPCC?). I already pushed a branch a few weeks ago that did the MIPS-like behaviour for CJALR, or thought I did but can't find it. The real work needs to be in ensuring xEPC/xEPCC no longer get legalised on _write_ but on _read_, since toggling misa.C changes the legalisation behaviour. If I write a 16-bit aligned PC, clear misa.C, read xEPC, set misa.C, read xEPC again, the spec requires PC[1] to be clear for the first read but set for the second. We don't currently get this correct, but it becomes worse in the presence of sentries, since it provides a construct to align a sentry down. Well, actually, we currently always tag-clear when writing sealed things to xEPCC, and that needs to be fixed too. But we need to do this refactoring before we can introduce sentries. Prashanth has a sail-riscv branch to do this, but I don't know what state it's in. We also need to ensure that the xepc extension hook in sail-cheri-riscv is able to reliably determine (ie is told by sail-riscv) whether the current xEPC is legal. There's also the difficult question of what to do when the base is only 16-bit aligned but you then turn C off. This can of worms is why I never finished the sentries work; modifying CJALR and friends was comparatively easy.

src/cheri_sys_exceptions.sail
nwf: Ah, fair.  I'll drop this hunk.

nwf:
I think I was confused and took `TCC` as `EPCC`, thus the code to unseal it.  I'll go have a look at the EPCC code.

arichardson:
Can we avoid some of this complexity by disallowing changes to misa.C?

> The Extensions field is a WARL field that can contain writable bits where the implementation allows the supported ISA to be modified.

jrtc27:
Sure, we _can_ impose such restrictions, but then we lose generality. Given the model is the specification we have deliberately been ensuring we describe the behaviour in the presence of all optional features of the specification. Not doing so would mean no CHERI implementation could support changing misa.C. It’s not _that_ difficult to do correctly, it’s just a bit of busy work that I didn’t have time to do given the state of all the F1 stuff.

src/cheri_insts.sail
```diff
@@ -666,19 +672,50 @@ function clause execute (CSeal(cd, cs, ct)) =
...
+      let (success, newCap) = sealCap(cs_val, to_bits(otype_width, ct_cursor));
+      if not (success) then {
+        handle_cheri_reg_exception(CapEx_InexactBounds, cs);
+        RETIRE_FAIL
+      } else {
+        writeCapReg(cd, cs_val);
+        RETIRE_SUCCESS
+      }
+    } else {
+      handle_cheri_reg_exception(CapEx_TagViolation, cs); // XXX?
```
PeterRugg, line 716: Shouldn't this fall-through case be writeCapReg(cd, cs_val); RETIRE_SUCCESS as before? Presumably this is just copied from CSeal, or is this an intended behaviour change?

src/cheri_insts.sail
```diff
@@ -666,19 +672,50 @@ function clause execute (CSeal(cd, cs, ct)) =
...
+     * unrepresentable.  In that case, the caller may hold another authority
+     * they wish to try, so just pass the result through.
+     */
+    if ct_cursor == otype_sentry then {
+      let (success, newCap) = sealCap(cs_val, to_bits(otype_width, ct_cursor));
+      if not (success) then {
+        handle_cheri_reg_exception(CapEx_InexactBounds, cs);
+        RETIRE_FAIL
+      } else {
+        writeCapReg(cd, cs_val);
```
PeterRugg, line 712: newCap rather than cs_val?

nwf: Yes, good find.

src/cheri_insts.sail
nwf: Ah, yes, right, this probably should just fall through as before.  Don't know how I got my wires crossed; maybe it was, indeed, copy-paste.  Will fix.

PeterRugg:
There seems to be an issue where cspecialrw into mepcc of a cap with an odd offset followed by an mret causes the new PC to have the LSB set. It looks like this is because prepare_xret_target doesn't legalize away the bottom bit (pc_alignment_mask appears to have the least significant bit set!?)

src/cheri_insts.sail
```diff
@@ -666,19 +672,50 @@ function clause execute (CSeal(cd, cs, ct)) =
...
     RETIRE_FAIL
-  } else if not (ct_val.tag) | ct_val.address == to_bits(sizeof(xlen), otype_unsealed) then {
-    writeCapReg(cd, cs_val);
-    RETIRE_SUCCESS
   } else if cs_val.sealed then {
+    /*
+     * If the input to be conditionally sealed is already sealed, just pass it
+     * through before any other checks are made.
+     */
     handle_cheri_reg_exception(CapEx_SealViolation, cs);
```
PeterRugg, line 688: Comment disagrees with the code here? It seems the comment is correct?

nwf: `CCSeal` has really not been my finest work, has it?  Sorry!

nwf:
> There seems to be an issue where cspecialrw into mepcc of a cap with an odd offset followed by an mret causes the new PC to have the LSB set. It looks like this is because prepare_xret_target doesn't legalize away the bottom bit (pc_alignment_mask appears to have the least significant bit set!?)

So, do we think that an *unsealed* cap like that should end up being installed into PCC with its bottom bit cleared, while a sentry cap like that should end up being installed into PCC with its tag cleared?

src/cheri_sys_regs.sail
```diff
@@ -93,6 +93,30 @@ function legalize_tcc(o : Capability, v : Capability) -> Capability = {
...
+ *  - misa.C is disabled and the EPCC offset is sufficiently aligned
+ *
+ * If misa.C is held constant, as it almost always will be, between fault and
+ * readback, ?EPCC will be faithfully read back, because the alignment will
+ * adhere to the requirements of the architecture, which is what legalization
+ * is intending to enforce.
+ *
+ * If neither of those conditions hold, the resulting readout will be a lie; if
+ * ?EPCC has additionally been set to be a sentry, then the result will be an
+ * untagged lie.
```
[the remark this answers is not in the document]
jrtc27: I think the conclusion was "don't do that" before. There's a base alignment exception for certain cases but in general if software does stupid things then it can pick up the pieces. You can currently have a misaligned base on MIPS and confuse things that look at PC.

jrtc27:
> > There seems to be an issue where cspecialrw into mepcc of a cap with an odd offset followed by an mret causes the new PC to have the LSB set. It looks like this is because prepare_xret_target doesn't legalize away the bottom bit (pc_alignment_mask appears to have the least significant bit set!?)
> 
> So, do we think that an _unsealed_ cap like that should end up being installed into PCC with its bottom bit cleared, while a sentry cap like that should end up being installed into PCC with its tag cleared?

Legalise _and_ tag clear.

nwf:
Whoops.  Does the one-line fix in 994bac67 fix the issue, @PeterRugg?

src/cheri_insts.sail
PeterRugg: No worries, my first implementations of features were no better before testing!

PeterRugg:
Was the intention to have cbuildcap preserve sentries?

nwf:
If memory serves, the reason we didn't have CBuildCap constructing sentries directly was that we wanted CBuildCap to also act as CUnsealViaSuperset even on sentries, and so the task of re-sealing as a sentry fell to CCSeal (and CCopyType).

nwf:
Well, if we've made CBuildCap the thing that seals sentries, then I think I should rip out the CCSeal special case for them.

PR #22, opened 2020-06-12
Add sail-coverage support.

pmundkur:
Uses a SAILCOV environment variable to build sail-coverage support.  It dumps the branch information into generated_definitions/c/all_branches.  This will require the coverage library to be built in the Sail compiler build, under sail/lib/coverage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,7 +292,7 @@ c_emulator/cheri_riscv_sim_%: generated_definitions/c/riscv_model_%.c $(SAIL_RIS
 
 generated_definitions/c/riscv_rvfi_model_%.c: $(SAIL_RVFI_SRCS) $(SAIL_RISCV_MODEL_DIR)/main.sail Makefile
 	mkdir -p generated_definitions/c
-	$(SAIL) $(SAIL_FLAGS) -O -Oconstant_fold -memo_z3 -c -c_include riscv_prelude.h -c_include riscv_platform.h -c_no_main $(SAIL_RVFI_SRCS) $(SAIL_RISCV_MODEL_DIR)/main.sail -o $(basename $@)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -275,7 +281,7 @@ generated_definitions/c/riscv.c: $(SAIL_SRCS) $(SAIL_RISCV_MODEL_DIR)/main.sail
 
 generated_definitions/c/riscv_model_%.c: $(SAIL_SRCS) $(SAIL_RISCV_MODEL_DIR)/main.sail Makefile
 	mkdir -p generated_definitions/c
-	$(SAIL) $(SAIL_FLAGS) -O -Oconstant_fold -memo_z3 -c -c_include riscv_prelude.h -c_include riscv_platform.h -c_no_main $(SAIL_SRCS) $(SAIL_RISCV_MODEL_DIR)/main.sail -o $(basename $@)
+	$(SAIL) $(SAIL_FLAGS) -O -Oconstant_fold -memo_z3 -c -c_include riscv_prelude.h -c_include riscv_platform.h -c_no_main $(SAIL_SRCS) $(SAIL_RISCV_MODEL_DIR)/main.sail -o $(basename $@) > generated_definitions/c/all_branches
```
jrtc27, line 284: This doesn't seem like a great interface from sail... stdout really isn't meant for this.

Alasdair: If you want to change it I don't mind. It was just the most braindead simple solution rather than passing an extra file handle to everywhere it might be needed.

Alasdair: Thinking about it changing `val branch_coverage : bool` into `val branch_coverage : out_channel option` in the ANF -> IR config module is probably the best way to go. I won't have time to do this before our PLDI workshop next week but it is probably a pretty easy change.

pmundkur:
the stdout issue is now fixed. any objections to merging this?

jrtc27:
No, other than please bump the sail-riscv submodule to include the change within it to use a file rather than stdout for coverage so everything is consistent.

PR #23, opened 2020-06-20
Fix PTE exceptions

jrtc27:
Things were broken in several ways:

1. Our SC page fault was not tag-dependent

2. We never actually got custom exceptions, just vanilla page faults for missing LoadCap/StoreCap bits

3. We weren't using the recently-added top-level exception codes (what I initially came to fix, before going down the rabbit hole of 1 and 2)

This fixes all of that in a far cleaner way than the highly-duplicated code that was there before. Some of this likely overlaps with Wes's more featureful changes, but we need to get this merged to be able to TestRIG anything with the CHERI PTE bits and those are more invasive. This should then provide a better basis for them to work off.

This depends on https://github.com/rems-project/sail-riscv/pull/62 being merged (and when it is I will update this PR to use the corresponding merge commit for the submodule bump).

Passes my own hand-written tests (magic macros, support code to set up paging and run the tests, and trap handler that makes traps look like writebacks to s6/s7, not shown):

```asm
/*
 * Global register allocations
 * s0 - current test
 * s1 - vatopa
 * s2 - active PTE address
 * s3 - clean PTE value
 * s4 - scratch page
 * s6 - last mcause
 * s7 - last mtval
 */

BEGIN_TEST
	/* Test that we can store and load back a tagged value */
	cspecialr	ct0, ddc
	/* Sanity check */
	cgettag	t1, ct0
	ASSERT_EQ_I	t1, 1, t6
	/* Round trip through memory */
	sc	ct0, 0(s4)
	lc	ct0, 0(s4)
	/* Assert nothing trapped */
	ASSERT_EQ_I	s6, ~0, t6
	/* Assert the tag is still there */
	cgettag	t1, ct0
	ASSERT_EQ_I	t1, 1, t6
END_TEST

BEGIN_TEST
	/* Test that dropping PTE_SC traps for tagged values */
	/* Update the PTE */
	li	t0, ~(PTE_SC)
	and	t0, s3, t0
	sd	t0, 0(s2)
	sfence.vma
	cspecialr	ct0, ddc
	/* Sanity check */
	cgettag	t1, ct0
	ASSERT_EQ_I	t1, 1, t6
	/* Do the store */
	sc	ct0, 0(s4)
	/* Assert we trapped */
	ASSERT_EQ_I	s6, EXC_SAMO_CAP_FAULT, t6
	/* Assert we saw the right faulting address */
	ASSERT_EQ	s7, s4
END_TEST

BEGIN_TEST
	/* Test that dropping PTE_SC doesn't trap for untagged values */
	/* Update the PTE */
	li	t0, ~(PTE_SC)
	and	t0, s3, t0
	sd	t0, 0(s2)
	sfence.vma
	/* Do the store */
	sc	cnull, 0(s4)
	/* Assert nothing trapped */
	ASSERT_EQ_I	s6, ~0, t6
END_TEST

BEGIN_TEST
	/* Test that dropping PTE_LC tag strips on load */
	cspecialr	ct0, ddc
	/* Sanity check */
	cgettag	t1, ct0
	ASSERT_EQ_I	t1, 1, t6
	/* Store tagged value to location */
	sc	ct0, 0(s4)
	/* Sanity check */
	lc	ct0, 0(s4)
	cgettag	t1, ct0
	ASSERT_EQ_I	t1, 1, t6
	/* Update the PTE */
	li	t0, ~(PTE_LC)
	and	t0, s3, t0
	sd	t0, 0(s2)
	sfence.vma
	/* Do the load */
	lc	ct0, 0(s4)
	/* Assert nothing trapped */
	ASSERT_EQ_I	s6, ~0, t6
	/* Assert the tag was stripped */
	cgettag	t1, ct0
	ASSERT_EQ_I	t1, 0, t6
END_TEST
```

### DIFF
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -1206,7 +1206,7 @@ function handle_store_cap_via_cap(rs, cs, cap_val, vaddrBits) = {
   } else if not(is_aligned_addr(vaddrBits, cap_size)) then {
     handle_mem_exception(vaddrBits, E_SAMO_Addr_Align());
     RETIRE_FAIL
-  } else match translateAddr(vaddrBits, Write(Cap)) {
+  } else match translateAddr(vaddrBits, Write(if rs_val.tag then Cap else Data)) {
     TR_Failure(E_Extension(EXC_CHERI_VMEM_STORE_CAP), _) => {
       handle_cheri_cap_exception(CapEx_TLBNoStoreCap, cs);
       RETIRE_FAIL
@@ -1354,7 +1354,7 @@ function handle_store_cond_cap_via_cap(cs2, cs, cap_val, vaddrBits) = {
     cancel_reservation();
     RETIRE_SUCCESS
   } else {
-    match translateAddr(vaddrBits, Write(Cap)) {
+    match translateAddr(vaddrBits, Write(if cs2_val.tag then Cap else Data)) {
       TR_Failure(E_Extension(EXC_CHERI_VMEM_STORE_CAP), _) => {
         handle_cheri_cap_exception(CapEx_TLBNoStoreCap, cs);
         RETIRE_FAIL
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -1207,11 +1207,6 @@ function handle_store_cap_via_cap(rs, cs, cap_val, vaddrBits) = {
     handle_mem_exception(vaddrBits, E_SAMO_Addr_Align());
     RETIRE_FAIL
   } else match translateAddr(vaddrBits, Write(if rs_val.tag then Cap else Data)) {
-    TR_Failure(E_Extension(EXC_CHERI_VMEM_STORE_CAP), _) => {
-      handle_cheri_cap_exception(CapEx_TLBNoStoreCap, cs);
-      RETIRE_FAIL
-    },
-    TR_Failure(E_Extension(_), _) => { internal_error("unexpected cheri exception for cap store") },
     TR_Failure(e, _) => { handle_mem_exception(vaddrBits, e); RETIRE_FAIL },
     TR_Address(addr, _) => {
       let eares : MemoryOpResult(unit) = mem_write_ea_cap(addr, aq, rl, false);
@@ -1355,11 +1350,6 @@ function handle_store_cond_cap_via_cap(cs2, cs, cap_val, vaddrBits) = {
     RETIRE_SUCCESS
   } else {
     match translateAddr(vaddrBits, Write(if cs2_val.tag then Cap else Data)) {
-      TR_Failure(E_Extension(EXC_CHERI_VMEM_STORE_CAP), _) => {
-        handle_cheri_cap_exception(CapEx_TLBNoStoreCap, cs);
-        RETIRE_FAIL
-      },
-      TR_Failure(E_Extension(_), _) => { internal_error("unexpected cheri exception for cap store") },
       TR_Failure(e, _) => { handle_mem_exception(vaddrBits, e); RETIRE_FAIL },
       TR_Address(addr, _) => {
         let eares : MemoryOpResult(unit) = mem_write_ea_cap(addr, aq, rl, false);
--- a/src/cheri_pte.sail
+++ b/src/cheri_pte.sail
@@ -42,98 +42,62 @@ union PTE_Check = {
   PTE_Check_Failure : ext_ptw
 }
 
-function to_pte_check(b : bool) -> (bool, pte_check) =
-  (b, PTE_CAP_OK)
-
 function checkPTEPermission(ac : AccessType(ext_access_type), priv : Privilege, mxr : bool, do_sum : bool, p : PTE_Bits, ext : extPte, ext_ptw : ext_ptw) -> PTE_Check = {
-  let (succ, pte_chk) : (bool, pte_check) =
+  /*
+   * Although in many cases MXR doesn't make sense for capabilities, we honour
+   * it for three reasons:
+   *
+   * 1. It provides uniformity rather than giving strange and surprising edge cases.
+   *
+   * 2. The tag-dependence of stores is achieved by passing Data in for untagged
+   *    capabilities. Thus, not honouring MXR for capabilities would result in
+   *    differences in whether MXR had an effect based on the tag.
+   *
+   * 3. It's simpler to implement yet still safe (LC is unaffected by MXR).
+   */
+  let base_succ : bool =
   match (ac, priv) {
-    (Read(Data),      User)       => to_pte_check(p.U() == 0b1 & (p.R() == 0b1 | (p.X() == 0b1 & mxr))),
-    (Write(Data),     User)       => to_pte_check(p.U() == 0b1 & p.W() == 0b1),
-    (ReadWrite(Data), User)       => to_pte_check(p.U() == 0b1 & p.W() == 0b1 & (p.R() == 0b1 | (p.X() == 0b1 & mxr))),
-    (Execute(),       User)       => to_pte_check(p.U() == 0b1 & p.X() == 0b1),
+    (Read(_),      User)       => p.U() == 0b1 & (p.R() == 0b1 | (p.X() == 0b1 & mxr)),
+    (Write(_),     User)       => p.U() == 0b1 & p.W() == 0b1,
+    (ReadWrite(_), User)       => p.U() == 0b1 & p.W() == 0b1 & (p.R() == 0b1 | (p.X() == 0b1 & mxr)),
+    (Execute(),    User)       => p.U() == 0b1 & p.X() == 0b1,
 
-    (Read(Data),      Supervisor) => to_pte_check((p.U() == 0b0 | do_sum) & (p.R() == 0b1 | (p.X() == 0b1 & mxr))),
-    (Write(Data),     Supervisor) => to_pte_check((p.U() == 0b0 | do_sum) & p.W() == 0b1),
-    (ReadWrite(Data), Supervisor) => to_pte_check((p.U() == 0b0 | do_sum) & p.W() == 0b1 & (p.R() == 0b1 | (p.X() == 0b1 & mxr))),
-    (Execute(),       Supervisor) => to_pte_check(p.U() == 0b0 & p.X() == 0b1),
+    (Read(_),      Supervisor) => (p.U() == 0b0 | do_sum) & (p.R() == 0b1 | (p.X() == 0b1 & mxr)),
+    (Write(_),     Supervisor) => (p.U() == 0b0 | do_sum) & p.W() == 0b1,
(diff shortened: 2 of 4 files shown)


src/cheri_pte.sail
```diff
@@ -42,98 +42,62 @@ union PTE_Check = {
...
+   * Although in many cases MXR doesn't make sense for capabilities, we honour
+   * it for three reasons:
+   *
+   * 1. It provides uniformity rather than giving strange and surprising edge cases.
+   *
+   * 2. The tag-dependence of loads is achieved by passing Data in for untagged
+   *    capabilities. Thus, not honouring MXR for capabilities would result in
+   *    differences in whether MXR had an effect based on the tag.
+   *
+   * 3. It's simpler to implement yet still safe (LC is unaffected by MXR).
```
jrtc27, line 56: I should probably change this to be clear that I mean the LC _bit_, because LC-the-instruction is definitely affected by MXR, that's what the whole discussion is about.

src/cheri_insts.sail
```diff
@@ -1206,12 +1206,7 @@ function handle_store_cap_via_cap(rs, cs, cap_val, vaddrBits) = {
   } else if not(is_aligned_addr(vaddrBits, cap_size)) then {
     handle_mem_exception(vaddrBits, E_SAMO_Addr_Align());
     RETIRE_FAIL
-  } else match translateAddr(vaddrBits, Write(Cap)) {
-    TR_Failure(E_Extension(EXC_CHERI_VMEM_STORE_CAP), _) => {
-      handle_cheri_cap_exception(CapEx_TLBNoStoreCap, cs);
-      RETIRE_FAIL
-    },
-    TR_Failure(E_Extension(_), _) => { internal_error("unexpected cheri exception for cap store") },
+  } else match translateAddr(vaddrBits, Write(if rs_val.tag then Cap else Data)) {
```
nwf, line 1209: I've left similar branching on `rs_val.tag` in #18, but the more I think about it the more I'm convinced it's not the correct approach.  It doesn't work for amoswap.c, as there's but one flag to convey both load and store.  This suggests that we should always be passing `Cap` in and that it's up to the PTW logic, not `cheri_insts`, to decide whether we're operating in a data-dependent way (by returning a caveated `TR_Address`) or -independent (by returning `TR_Failure`).

As we don't have amoswap.c yet and I don't wish to impede progress, I'm happy for this to land as it stands and we can shake out the details as part of #18.

src/cheri_riscv_types.sail
```diff
@@ -24,27 +24,46 @@ function ext_accum_ptw_result(ptw : ext_ptw, pte : pte_check) -> ext_ptw =
...
 
 /* CHERI conversion of extension exceptions to integers */
 val num_of_ext_exc_type : ext_exc_type -> {'n, (0 <= 'n < xlen). int('n)}
-function num_of_ext_exc_type(e) = 28
+function num_of_ext_exc_type(e) =
+  match (e) {
+    EXC_LOAD_CAP_PAGE_FAULT => 26,
+    EXC_SAMO_CAP_PAGE_FAULT => 27,
+    EXC_CHERI               => 28
+  }
```
nwf, line 60: I can't help but feel that one of `num_of_ext_exc_type` and `ext_exc_type_to_bits` should be redundant.  (This presumably requires a change in `riscv-sail` as well).

src/cheri_ptw.sail
```diff
@@ -24,14 +24,21 @@ function ptw_error_to_str(e) =
 
 overload to_str = {ptw_error_to_str}
 
+function ext_get_ptw_error(ext_ptw : ext_ptw) -> PTW_Error =
+  match (ext_ptw) {
+    PTW_CAP_OK        => PTW_No_Permission(),
+    PTW_LOAD_CAP_ERR  => PTW_Ext_Error(AT_CAP_ERR),
+    PTW_STORE_CAP_ERR => PTW_Ext_Error(AT_CAP_ERR)
```
nwf, line 31: Why have two PTW errors, conflate them down to one PTW_Ext_Error, and then pry them back apart in `translationException`?

src/cheri_pte.sail
```diff
@@ -42,98 +42,62 @@ union PTE_Check = {
...
-
 function checkPTEPermission(ac : AccessType(ext_access_type), priv : Privilege, mxr : bool, do_sum : bool, p : PTE_Bits, ext : extPte, ext_ptw : ext_ptw) -> PTE_Check = {
-  let (succ, pte_chk) : (bool, pte_check) =
+  /*
+   * Although in many cases MXR doesn't make sense for capabilities, we honour
+   * it for three reasons:
+   *
+   * 1. It provides uniformity rather than giving strange and surprising edge cases.
+   *
+   * 2. The tag-dependence of loads is achieved by passing Data in for untagged
```
nwf, line 52: Come again?

src/cheri_pte.sail
```diff
@@ -42,98 +42,62 @@ union PTE_Check = {
...
+    (Execute(),    Supervisor) => p.U() == 0b0 & p.X() == 0b1,
+
+    (_,            Machine)    => internal_error("m-mode mem perm check")
+  };
+
+  let e = Mk_Ext_PTE_Bits(ext);
+  let (succ, pte_chk) : (bool, pte_check) =
+  match (base_succ, ac) {
+    /* Base translation exceptions take priority over CHERI exceptions */
+    (false, _)               => (false, PTE_CAP_OK),
```
nwf, line 77: Possibly stylistically cleaner as
```
let (succ, pte_check) : (bool, pte_check) =
  if not (base_succ) (false, PTE_CAP_OK)
  else match ac {
    ...
  }
```

src/cheri_pte.sail
```diff
@@ -42,98 +42,62 @@ union PTE_Check = {
...
+
+    (true,  Read(Cap))       => if   e.LoadCap() == 0b1
+                                then (true, PTE_CAP_OK)
+                                /* Allow the address translation to proceed, but mark for tag stripping */
+                                else (true, PTE_LOAD_CAP_ERR),
+
+    (true,  Write(Cap))      => if   e.StoreCap() == 0b1
+                                then (true, PTE_CAP_OK)
+                                /* Do not allow the address translation to proceed */
+                                else (false, PTE_STORE_CAP_ERR),
```
nwf, line 87: #18 considered this OK and left the caveating to `cheri_insts` for data-dependence's sake.

src/cheri_pte.sail
```diff
@@ -42,98 +42,62 @@ union PTE_Check = {
...
+
+    (true,  ReadWrite(Cap))  => if   e.StoreCap() == 0b1 & e.LoadCap() == 0b1
+                                then (true, PTE_CAP_OK)
+                                else if e.StoreCap() == 0b0
+                                /* Do not allow the address translation to proceed */
+                                then (false, PTE_STORE_CAP_ERR)
```
nwf, line 93: Again #18 left this OK.

src/cheri_insts.sail
jrtc27: Yeah, that will need more work. This was just the minimal set of things needed to get it working today rather than in a week or month.

src/cheri_riscv_types.sail
jrtc27: I had the same thought, it's absurd. But yes, sail-riscv needs to change for that.

src/cheri_pte.sail
jrtc27: .. stores, yes

src/cheri_ptw.sail
jrtc27: You're right, currently we should have just the one PTE_CAP_ERR/PTW_CAP_ERR, that's more in keeping with the base model. But AMOSWAP.C changes that if we go with your suggestion of returning _true_ for SC violations and letting them through, since then the code needs to be able to distinguish between LC and SC being missing so it knows whether to trap or tag clear. Thus I think this is right; we keep the state about the details of the page table walk in ext_ptw, and when we need to trap we report a generic trap that's then turned into the right one based on the _instruction_ (because AMOs always trap with store/AMO faults, even if it was the load part of it that caused the trap).

src/cheri_insts.sail
jrtc27: Also, this approach _would_ work if we made ReadWrite take a pair; then you could have Cap always in the first and Data/Cap in the second for AMOSWAP.C. That then pushes the trapping logic into the PTW where it really belongs, leaving just tag clearing to the instruction. I think that's actually nicest?

src/cheri_ptw.sail
nwf: > even if it was the load part of it that caused the trap

That's... not going to be fantastic for the load generation counter story (though of course we can hook both 0x1A and 0x1B in software for CLG checks).  I'd prefer we deviated from upstream here, but I'm open to more authoritative opinions.

jrtc27: I assume the reasoning is that your store fault handler normally does a strict superset of what your load fault handler does, since if you can store to it you can read from it (W-but-not-R is a reserved state I believe), so you might as well take the full fault up-front and do everything in one go rather than do the bare minimum for a load and then take store fault upon retry. My view is we should follow upstream, and then if it proves to be as painful as we fear for revocation, work out what the right approach is, but only then. It might turn out to not be expensive in practice.

src/cheri_insts.sail
nwf: That's what I was proposing in https://github.com/rems-project/sail-riscv/pull/57

jrtc27: I think I may have initially doubted that approach, but yes. Though we only want a subset of that; the r and w _types_ can be the same, just that you get two values in ReadWrite.

src/cheri_ptw.sail
nwf: Yea, in practice we shouldn't be doing too many AMOSWAP.C-s against CLG-mismatching pages, one hopes, and adding a test for CLG-mismatch to the SAMO handler should be cheap because SAMO's on the way to CoW most of the time, right?

src/cheri_riscv_types.sail
```diff
@@ -24,27 +24,46 @@ function ext_accum_ptw_result(ptw : ext_ptw, pte : pte_check) -> ext_ptw =
...
 
 /* CHERI translation of PTW errors into exception annotations */
 function ext_translate_exception(e : ext_ptw_error) -> ext_exc_type =
   match e {
-    AT_LOAD_CAP_ERR  => EXC_CHERI_VMEM_LOAD_CAP,
-    AT_STORE_CAP_ERR => EXC_CHERI_VMEM_STORE_CAP
-}
+    AT_LOAD_CAP_ERR  => EXC_LOAD_CAP_PAGE_FAULT,
+    AT_STORE_CAP_ERR => EXC_SAMO_CAP_PAGE_FAULT
+  }
```
nwf, line 42: This hunk seems stale -- I think `ext_translate_exception` is not used any more as of this commit and should just be removed.  (It's only called from within `riscv_ptw.sail` upstream, a file we override).

jrtc27: I view it as a bug that we have to do so. But yes, this hook doesn't make sense in its current form and should just go.

PR #24, opened 2020-07-20
Propose C{Get,Set}Bits for extended register access

nwf:
One possibility for https://github.com/CTSRD-CHERI/cheri-architecture/issues/35


arichardson:
I like this approach since it's flexible. In a language runtime it would make it easy to have a type field as the lowest byte that you can check with and(immediate) and then get bits 72-8 that contain the interesting data.

Should it wrap around/zero-extend/sign-extend for selector > 8?

src/cheri_insts.sail
```diff
@@ -1569,6 +1610,9 @@ mapping clause assembly = CSpecialRW(cd, cs, idx) <-> "CSpecialRW"  ^ spc() ^ re
 mapping clause assembly = CIncOffsetImmediate(cd, cb, imm12) <-> "CIncOffsetImmediate" ^ spc() ^ reg_name(cd) ^ sep() ^ reg_name(cb) ^ sep() ^ hex_bits_12(imm12)
 mapping clause assembly = CSetBoundsImmediate(cd, cb, imm12) <-> "CSetBoundsImmediate" ^ spc() ^ reg_name(cd) ^ sep() ^ reg_name(cb) ^ sep() ^ hex_bits_12(imm12)
 
+mapping clause assembly = CGetBits(rd, cs, sel) <-> "CGetBits" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(cs) ^ sep() ^ hex_bits_5(sel)
+mapping clause assembly = CSetBits(cd, rs, sel) <-> "CGetBits" ^ spc() ^ reg_name(cd) ^ sep() ^ reg_name(rs) ^ sep() ^ hex_bits_5(sel)
```
arichardson, line 1614: CSetBits

jrtc27:
IMO you should trap if you ever read past the end. Those bits are meaningless and, whilst perhaps harmless for CGetBits, black-holing them on CSetBits would risk hiding software bugs and I see no good reason to allow that, so might as well also trap symmetrically for CGetBits.

arichardson:
Sign extending could be useful if you wanted to do a union holding a 120-bit integer without additional shifts. But trapping is also fine.

I think we should definitely trap for all bits outside the capability though.

nwf:
Here's a version that adds sign extension to CGetBits.  Entirely-OOB accesses in CGetBits trap, and Any-OOB accesses in CSetBits trap.  The asymmetry is unfortunate, but it seemed better than letting some bits land on the floor.

rwatson:
It would be worth thinking about and documenting potential semantics that are trap-free, as this is essentially an arithmetic operation, and would be used frequently in fast paths. Arm, for example, would likely not accept a trapping version.

arichardson:
> It would be worth thinking about and documenting potential semantics that are trap-free, as this is essentially an arithmetic operation, and would be used frequently in fast paths. Arm, for example, would likely not accept a trapping version.

It's probably easier in this case since you can tell during decode? All CGetBits shift values 15 are invalid opcodes and all CSetBits values > 8.

PR #25, opened 2020-08-10
Add CSetEqualExact

nwf:
While we generally compare capabilities by virtual address or through other
explicit projections, sometimes (e.g., the CAS involved in revocation) we
really do mean bit-for-bit equality of the encodings.

### DIFF
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -574,6 +574,15 @@ function clause execute (CTestSubset(rd, cb, ct)) =
   RETIRE_SUCCESS
 }
 
+union clause ast = CSEQX : (regidx, regidx, regidx)
+function clause execute (CSEQX(rd, cb, ct)) =
+{
+  let cb_val = readCapReg(cb);
+  let ct_val = readCapReg(ct);
+  X(rd) = EXTZ(bool_to_bits(cb_val == ct_val));
+  RETIRE_SUCCESS
+}
+
   /*
    * Common suffix of CSeal and CCSeal
    */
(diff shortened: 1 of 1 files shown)


src/cheri_insts.sail
```diff
@@ -574,6 +574,15 @@ function clause execute (CTestSubset(rd, cb, ct)) =
   RETIRE_SUCCESS
 }
 
+union clause ast = CSetEqualExact : (regidx, regidx, regidx)
+function clause execute (CSetEqualExact(rd, cb, ct)) =
+{
+  let cb_val = readCapRegDDC(cb);
```
arichardson, line 580: As I mentioned in https://github.com/CTSRD-CHERI/qemu/pull/121, we may want to use CNULL?

src/cheri_insts.sail
```diff
@@ -1516,6 +1527,8 @@ mapping clause assembly = CSpecialRW(cd, cs, idx) <-> "CSpecialRW"  ^ spc() ^ re
 mapping clause assembly = CIncOffsetImmediate(cd, cb, imm12) <-> "CIncOffsetImmediate" ^ spc() ^ reg_name(cd) ^ sep() ^ reg_name(cb) ^ sep() ^ hex_bits_12(imm12)
 mapping clause assembly = CSetBoundsImmediate(cd, cb, imm12) <-> "CSetBoundsImmediate" ^ spc() ^ reg_name(cd) ^ sep() ^ reg_name(cb) ^ sep() ^ hex_bits_12(imm12)
 
+mapping clause assembly = CSetEqualExact(rd, cb, ct) <-> "CSetEqualExact" ^ spc() ^ reg_name(rd) ^ sep() ^ reg_name(cb) ^ sep() ^ reg_name(ct)
```
jrtc27, line 1530: CSEQX

src/cheri_insts.sail
```diff
@@ -574,6 +574,15 @@ function clause execute (CTestSubset(rd, cb, ct)) =
   RETIRE_SUCCESS
 }
 
+union clause ast = CSetEqualExact : (regidx, regidx, regidx)
+function clause execute (CSetEqualExact(rd, cb, ct)) =
+{
+  let cb_val = readCapReg(cb);
+  let ct_val = readCapReg(ct);
+  X(rd) = EXTZ(if cb_val == ct_val then 0b1 else 0b0);
```
jrtc27, line 582:
```suggestion
  X(rd) = EXTZ(bool_to_bits(cb_val == ct_val));
```

arichardson:
@jrtc27 can we merge this?

jrtc27:
Oh this is still open. I wanted union tag called CSEQX, not just the mnemonic.

arichardson:
> Oh this is still open. I wanted union tag called CSEQX, not just the mnemonic.

Personally I prefer the longer name (since it makes it obvious what the instruction does without having to look at the definition) but don't mind either way.

jrtc27:
Hm, I thought it was CSEQX in the spec but it's the full thing there. In that case, I should update the spec :)

I agree it makes it more obvious what the instruction does, but that's not the RISC-V way...

arichardson:
I agree it's not the RISC-V way, but doing the RISC-V way for just one of the instructions is also inconsistent. I guess if we ever want RISC-V folks to adopt CHERI we need to change CGetBase -> cgb, CGetOffset -> cgo, CIncOffset -> cinc, etc.

jrtc27:
The difference is that CGetOffset has nothing like it in baseline RISC-V, but CSEQX has all kinds of equivalents. Same as how we do CLW not CapLoadWord.

jrtc27:
Also the MIPS one is CEXEQ so CSetEqualExact is inconsistent there :P

arichardson:
> The difference is that CGetOffset has nothing like it in baseline RISC-V, but CSEQX has all kinds of equivalents. Same as how we do CLW not CapLoadWord.

True, but then CIncOffset should probably be CADD/CADDI since it's a capability add.

arichardson (approved):

PR #27, opened 2020-09-25
Add saildoc for CAndPerm

arichardson:
This is the existing MIPS description with the register names updated and
converted to markdown.

If this seems reasonable, I'll move over the other descriptions.

### DIFF
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -189,6 +189,19 @@ function clause execute (CSpecialRW(cd, scr, cs1)) =
 }
 
 union clause ast = CAndPerm : (regidx, regidx, regidx)
+/*!
+ * Capability register *cd* is replaced with the contents of capability
+ * register *cs1* with the **perms** field set to the bitwise and of its
+ * previous value and bits 0 to [cap_hperms_width]-1 of integer register *rs2*
+ * and the **uperms** field set to the bitwise and of its previous value and
+ * bits [cap_uperms_shift] to [cap_uperms_shift]+[cap_uperms_width]-1 of *rd*.
+ *
+ *
+ * ## Exceptions:
+ * An exception is raised if:
+ *   - *cs1.tag* is not set.
+ *   - *cs1* is sealed.
+ */
 function clause execute(CAndPerm(cd, cs1, rs2)) =
 {
   let cs1_val = C(cs1);
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -196,8 +196,8 @@ union clause ast = CAndPerm : (regidx, regidx, regidx)
  * and the **uperms** field set to the bitwise and of its previous value and
  * bits [cap_uperms_shift] to [cap_uperms_shift]+[cap_uperms_width]-1 of *rd*.
  *
+ * ## Exceptions
  *
- * ## Exceptions:
  * An exception is raised if:
  *   - *cs1.tag* is not set.
  *   - *cs1* is sealed.
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -199,7 +199,7 @@ union clause ast = CAndPerm : (regidx, regidx, regidx)
  * ## Exceptions
  *
  * An exception is raised if:
- *   - *cs1.tag* is not set.
+ *   - *cs1*.**tag** is not set.
  *   - *cs1* is sealed.
  */
 function clause execute(CAndPerm(cd, cs1, rs2)) =


src/cheri_insts.sail
```diff
@@ -189,6 +189,13 @@ function clause execute (CSpecialRW(cd, scr, cs1)) =
 }
 
 union clause ast = CAndPerm : (regidx, regidx, regidx)
+/*!
+ * Capability register *cd* is replaced with the contents of capability
+ * register *cs1* with the \cperms{} field set to the bitwise-AND of
+ * its previous value and bits 0 .. 10 of integer register *rs2*
```
jrtc27, line 195: This isn't actually true any more; CID is bit 11 (and ASR is bit 10, unclear if `0 .. 10` includes 10 too or not, but probably?).

jrtc27:
Probably also shouldn't be using `..` given that will typeset strangely in TeX. Use words, `\ldots{}` (which will give 3 dots :/) or something else appropriate?

jrtc27:
Also needs a `\begin{exceptions} ... \end{exceptions}` bit to document the tag and seal violations like MIPS.

src/cheri_insts.sail
arichardson: We should probably use a macro to avoid bugs like this.

jrtc27: Now that the documentation is part of the Sail I think it's fine.

arichardson:
> Probably also shouldn't be using `..` given that will typeset strangely in TeX. Use words, `\ldots{}` (which will give 3 dots :/) or something else appropriate?

Yes that sounds good, I just used the existing format and did a naive conversion.

src/cheri_insts.sail
nwf: Maybe best to say something like `the \cperms{} and \cuperms{} fields bitwise-ANDed with their joint architectural representation (see \cref{..., ...}) in integer register *rs2*`?

src/cheri_insts.sail
```diff
@@ -189,6 +189,13 @@ function clause execute (CSpecialRW(cd, scr, cs1)) =
...
+ * and the \cuperms{} field set to the bitwise and of its previous value
```
nwf, line 196: At the very least, be consistent between `bitwise-AND` and `bitwise and`

src/cheri_insts.sail
jrtc27: As I discussed with Alex on Slack, these docs need to make sense standalone, so no `\cref`'s to cheri-architecture places (especially since they're prone to being renamed without realising this other repository needs to be updated).

src/cheri_insts.sail
arichardson: I hope it wasn't me who added that to the MIPS docs 🏃‍♂️

src/cheri_insts.sail
nwf: That's fair... but can we not cross-reference saildocs?

jrtc27: Currently cheri-architecture has no label for the documentation subsection, only one for the function clause itself and one for the whole instruction, so you'd (a) want to reference one of those two (b) use a suitable macro to abstract cheri-architecture specifics away.

nwf: Well, I'd meant more like pointing at the sail helpers `getCapPerms` and `setCapPerms` but while that's _named_ in 7.3 of the ISA doc, there's no analogous point in chapter 8, and even in 7.3 it's just a name, not the actual layout.

arichardson:
Now completely in markdown.

nwf (approved):

src/cheri_insts.sail
```diff
@@ -189,6 +189,19 @@ function clause execute (CSpecialRW(cd, scr, cs1)) =
...
 union clause ast = CAndPerm : (regidx, regidx, regidx)
+/*!
+ * Capability register *cd* is replaced with the contents of capability
+ * register *cs1* with the **perms** field set to the bitwise and of its
+ * previous value and bits 0 to [cap_hperms_width]-1 of integer register *rs2*
+ * and the **uperms** field set to the bitwise and of its previous value and
+ * bits [cap_uperms_shift] to [cap_uperms_shift]+[cap_uperms_width]-1 of *rd*.
+ *
+ *
+ * ## Exceptions:
```
jrtc27, line 200:
```suggestion
 *
 * ## Exceptions
 *
```

looks nicer I think (whitespace) and section headings shouldn't have colons in them.

src/cheri_insts.sail
```diff
@@ -189,6 +189,19 @@ function clause execute (CSpecialRW(cd, scr, cs1)) =
...
+ * An exception is raised if:
+ *   - *cs1.tag* is not set.
```
jrtc27, line 202: In two minds about whether this should be:
```suggestion
 *   - *cs1*.tag is not set.
```
or
```suggestion
 *   - *cs1*.**tag** is not set.
```
instead of what's there. (On the grounds that `cs1` is specifically referencing an operand named above, whereas `tag` is a property)

arichardson: `*cs1*.**tag** is not set.` matches MIPS, so I'll do that.

jrtc27: (the latter being the style currently used for MIPS)

jrtc27 (approved):
I'll work on a less-hacky cheri-architecture in parallel with your writing of other docs, but I think we can merge this even if cheri-architecture isn't quite ready (without a sed hack).

PR #28, opened 2020-09-28
Replace more hardcoded constants in the permissions interface

arichardson:


jrtc27:
If you're making those changes, please also add cap_ prefixes to otype_width and mantissa_width, otherwise things still aren't uniform. Also I'd prefer hperms to match the length of uperms, especially since it's getCapHardPerms not getCapHardwarePerms.

arichardson:
> If you're making those changes, please also add cap_ prefixes to otype_width and mantissa_width, otherwise things still aren't uniform. Also I'd prefer hperms to match the length of uperms, especially since it's getCapHardPerms not getCapHardwarePerms.

I was going to, but since there are more uses of those constants, I wanted to check that you aren't opposed to the change before making it.

jrtc27:
Nope, I was mildly frustrated by the inconsistency when I came to add cap_perms_width and went back and forth on whether to name it like the surrounding context or like the related uperms_width, so uniformity is a :+1: from me.

PR #29, opened 2020-09-29
Document CGet* instructions

arichardson:
These descriptions are based on the existing MIPS prose (converted to
markdown) with a few minor consistency tweaks.

### DIFF
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -44,6 +44,13 @@ union clause ast = CGetOffset : (regidx, regidx)
 union clause ast = CGetAddr   : (regidx, regidx)
 union clause ast = CGetFlags  : (regidx, regidx)
 
+/*!
+ * The least significant [cap_hperms_width] bits of integer register *rd* are
+ * set equal to the **perms** field of capability register *cs1*; bits
+ * [cap_uperms_shift] to [cap_uperms_shift]+[cap_uperms_width]-1 of *rd* are set
+ * equal to the **uperms** field of *cs1*.
+ * The other bits of \emph{rd} are set to zero.
+ */
 function clause execute (CGetPerm(rd, cs1)) =
 {
   let capVal = C(cs1);
@@ -51,6 +58,10 @@ function clause execute (CGetPerm(rd, cs1)) =
   RETIRE_SUCCESS
 }
 
+/*!
+ * Integer register *rd* is set equal to the zero-extended **flags** field of
+ * capability register *cs1*.
+ */
 function clause execute (CGetFlags(rd, cs1)) =
 {
   let capVal = C(cs1);
@@ -58,6 +69,10 @@ function clause execute (CGetFlags(rd, cs1)) =
   RETIRE_SUCCESS
 }
 
+/*!
+ * Integer register *rd* is set equal to the **otype** field of capability
+ * register *cs1*.
+ */
 function clause execute (CGetType(rd, cs1)) =
 {
   let capVal = C(cs1);
@@ -67,6 +82,10 @@ function clause execute (CGetType(rd, cs1)) =
   RETIRE_SUCCESS
 }
 
+/*!
+ * Integer register *rd* is set equal to the **base** field of capability
+ * register *cs1*.
+ */
 function clause execute (CGetBase(rd, cs1)) =
 {
   let capVal = C(cs1);
@@ -74,6 +93,10 @@ function clause execute (CGetBase(rd, cs1)) =
   RETIRE_SUCCESS
 }
 
+/*!
+ * Integer register *rd* is set equal to the **offset** field of capability
+ * register *cs1*.
+ */
 function clause execute (CGetOffset(rd, cs1)) =
 {
   let capVal = C(cs1);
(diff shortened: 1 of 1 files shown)


src/cheri_insts.sail
```diff
@@ -44,20 +44,36 @@ union clause ast = CGetOffset : (regidx, regidx)
...
 {
   let capVal = C(cs1);
   X(rd) = EXTZ(getCapFlags(capVal));
   RETIRE_SUCCESS
 }
 
+/*!
+ * Integer register *rd* is set equal to the sign-extended **otype** field
+ * of capability register *cs1*.
+ *
```
jrtc27, line 75:
```suggestion
```

src/cheri_insts.sail
```diff
@@ -67,20 +83,41 @@ function clause execute (CGetType(rd, cs1)) =
...
+ * inside math mode.
+ */
+/*!
+ * Integer register *rd* is set equal to the **length** field of capability
+ * register *cs1*.
+ *
+ * ## Notes
+ * Due to the compressed representation of capabilities, the actual length
+ * of capabilities can be $2^{[{xlen}][xlen]}$; [CGetLen] will return the maximum
+ * value of $2^{[{xlen}][xlen]}-1$ in this case.
```
jrtc27, line 119: I'm unsure how to deal with notes. Some of them make sense to keep in Sail, but some of the MIPS ones refer to uarch things and don't make so much sense there.

src/cheri_insts.sail
```diff
@@ -89,20 +126,33 @@ function clause execute (CGetLen(rd, cs1)) =
...
+ */
 function clause execute (CGetSealed(rd, cs1)) =
 {
   let capVal = C(cs1);
   X(rd) = EXTZ(bool_to_bits(capVal.sealed));
   RETIRE_SUCCESS
 }
 
+/*!
+ * Integer register *rd* is set equal to the **cursor** field (the address) of
```
jrtc27, line 153: The field is called address in RISC-V.

src/cheri_insts.sail
```diff
@@ -89,20 +126,33 @@ function clause execute (CGetLen(rd, cs1)) =
   RETIRE_SUCCESS
 }
 
+/*!
+ * The low bit of integer register *rd* is set to the **tag** value of *cs1*.
```
jrtc27, line 130: "tag value of cs1" reads strangely; use field instead like the others?

src/cheri_insts.sail
arichardson: I dropped some of the notes that don't make sense for RISC-V, but I felt like this one might make sense here?

src/cheri_insts.sail
arichardson: Sounds good, will update the MIPS description too.

src/cheri_insts.sail
arichardson: We still use `getCapCursor`, so until that is changed we should maybe match the MIPS description?

jrtc27: Those 3 instances should just be replaced with `cs2.address` like everywhere else and the function removed.

src/cheri_insts.sail
jrtc27: The issue though is what to do if we decide we do have uarch-related notes we want to include.

arichardson: How about: uarch notes that affect the architectural instruction semantics go into the sail, everything else into the uarch chapter of the ISA ref?

src/cheri_insts.sail
```diff
@@ -44,20 +44,35 @@ union clause ast = CGetOffset : (regidx, regidx)
...
 function clause execute (CGetFlags(rd, cs1)) =
 {
   let capVal = C(cs1);
   X(rd) = EXTZ(getCapFlags(capVal));
   RETIRE_SUCCESS
 }
 
+/*!
+ * Integer register *rd* is set equal to the sign-extended **otype** field
+ * of capability register *cs1*.
```
jrtc27, line 74: That's not quite true; normal otypes are positive, reserved otypes are negative, and the split is asymmetric.

jrtc27: And arguably like bounds the otype is architecturally an XLEN-bit value with restricted values due to architectural compression...

arichardson: MIPS has this:
```\subsection*{Notes}

\begin{itemize}
\item
If the capability is unsealed, a value of -1 is returned.
For hardware-interpreted \cotype{}s (see \cref{tab:archotypes}) a sign-extended
(negative) value is returned; for software-defined \cotype{}s
(see \cref{sec:model-sealedcapabilities}) \insnmipsref{CGetType} returns
a zero-extended value.
\end{itemize}
```

src/cheri_insts.sail
jrtc27: That seems sensible; we'll see whether there are any notes that don't fall into either category nicely.

src/cheri_insts.sail
```diff
@@ -67,20 +82,41 @@ function clause execute (CGetType(rd, cs1)) =
...
 
+/*
+ * Note: We have to use [{xlen}][xlen] instead of [xlen] to avoid \lstinline{}
+ * inside math mode.
+ */
+/*!
+ * Integer register *rd* is set equal to the **length** field of capability
+ * register *cs1*.
+ *
+ * ## Notes
```
jrtc27, line 115:
```suggestion
 * ## Notes
 *
```

src/cheri_insts.sail
arichardson: I could just drop the sign-extended and people can look at the sail to figure out the value?

jrtc27: Yeah that might be easiest, though we could add something reminding people that software-defined otypes are positive and architecturally-reserved otypes are negative.

jrtc27: Uh, under Notes, that is.

PR #30, opened 2020-09-30
Add basic documentation for CRRL, CRAM, CSub, CMove, CClearTag

arichardson:


src/cheri_insts.sail
```diff
@@ -309,6 +309,10 @@ function clause execute(CToPtr(rd, cs1, cs2)) =
 }
 
 union clause ast = CSub : (regidx, regidx, regidx)
+/*!
+ * Integer register *rd* is set equal to (*cs1*.**address** $-$
+ * *cs1*.**address**) $\bmod~2^{[{xlen}][xlen]}$.
```
PeterRugg, line 314:
```suggestion
 * *cs2*.**address**) $\bmod~2^{[{xlen}][xlen]}$.
```

src/cheri_insts.sail
```diff
@@ -620,8 +631,12 @@ function clause execute (CCopyType(cd, cs1, cs2)) =
   }
 }
 
-
 union clause ast = CRRL : (regidx, regidx)
+/*!
+ * Integer register *rt* is set to the smallest value greater or equal to *rs1*
```
PeterRugg, line 636:
```suggestion
 * Integer register *rd* is set to the smallest value greater or equal to *rs1*
```

src/cheri_insts.sail
```diff
@@ -620,8 +631,12 @@ function clause execute (CCopyType(cd, cs1, cs2)) =
...
+ * that can be used to set precise bounds on a capability that has a suitably
```
PeterRugg, line 637:
```suggestion
 * that can be used as a length to set precise bounds on a capability that has a suitably
```

src/cheri_insts.sail
```diff
@@ -630,6 +645,11 @@ function clause execute(CRRL(rd, rs1)) =
 }
 
 union clause ast = CRAM : (regidx, regidx)
+/*!
+ * Integer register *rt* is set to a mask that can be used to round addresses
```
PeterRugg, line 649:
```suggestion
 * Integer register *rd* is set to a mask that can be used to round addresses
```

PeterRugg:
```suggestion
 * Integer register *rt* is set to the most permissive mask that can be used to round addresses
```
Assuming this is in fact true...

PeterRugg (changes requested):

src/cheri_insts.sail
arichardson: I am not sure this is strictly the most permissive mask. I think CRAM right now is precise?

PeterRugg: Hmm, I'm not sure: it's based on CSetBounds which should be precise. We should probably leave it. Just noticed the asymmetry with CRRL's "smallest value greater or equal".

PeterRugg (approved):

src/cheri_insts.sail
arichardson: True, smallest in CRRL may also be an overstatement. But if I use "a value" there, then returning UINT64_MAX is a valid implementation.

PeterRugg: Just as returning 0 here is a valid implementation.

jrtc27: UINT64_MAX isn't representable, even if some existing capabilities claim to have that length :)

PeterRugg: Hmm, that's actually a good point. I recall conversations about this previously, but is CRRL specified to give UINT64_MAX in that case, or 0?

PR #31, opened 2020-09-30
CSetflags cleanup and documentation

arichardson:


src/cheri_insts.sail
```diff
@@ -272,6 +272,17 @@ function clause execute(CAndPerm(cd, cs1, rs2)) =
...
+/*!
+ * Capability register *cd* is replaced with the contents of capability
+ * register *cs1* with the **flags** field set to bits 0 to [cap_flags_width]-1
+ * of integer register *rs2*.
+ *
+ * ## Exceptions
+ *
+ * An exception is raised if:
+ *   - *cs1*.**tag** is not set.
+ *   - *cs1* is sealed.
```
jrtc27, line 284: Interestingly not; unlike some of the others, CSetFlags currently works on untagged things.

arichardson: copied from the MIPS description without looking at the implementation 🙈

arichardson: But it shouldn't do that for MIPS either.

PeterRugg (approved):

PR #32, opened 2020-09-30
Add documentation for the address-modification instructions

arichardson:


src/cheri_insts.sail
```diff
@@ -350,6 +363,19 @@ function clause execute (CIncOffset(cd, cs1, rs2)) =
 }
 
 union clause ast = CIncOffsetImmediate : (regidx, regidx, bits(12))
+/*!
+ * Capability register *cd* is set equal to capability register *cs1* with its
+ * **address** replaced with *cs1*.**address** $+$ *imm*.
+ * If the resulting capability cannot be represented exactly, then *cd*.**tag**
+ * is cleared. The remaining capability fields are set to what the in-memory
+ * representation of *cs1* with the address set to *cs1*.**address** $+$ *imm*
+ * decodes to.
```
jrtc27, line 372: We probably want to highlight that `imm` is signed, as for CSetBoundsImmediate it's unsigned?

arichardson: The RISC-V spec seems to use *imm* for signed, *uimm* for unsigned and *nzuimm* probably for non-zero unsigned?
How about we change CSetBoundsImm to use uimm?

jrtc27: Sure.

(nzuimm in RVC is, uh, "it depends": sometimes zero is illegal/a completely different instruction, sometimes it's "zero would be a kind of NOP so it's a compressed hint")

jrtc27: .. and we should update the ISA spec to call these `imm` and `uimm` rather than `offset` and `length`, which was always a strange choice.

jrtc27: Done in CTSRD-CHERI/cheri-architecture@16718c90.

jrtc27 (approved):

PR #33, opened 2020-10-03
Treat CCSeal with an out-of-bounds capability as a move

jrtc27:
We currently make sure to handle authorities that went unrepresentable due to a CCopyType, but not those that are still representable yet out of bounds.

Whilst here, reorder the CCSeal checks to match CSeal more (it's a series of branches all doing the same thing, so there's no functional difference) and document both CSeal and CCSeal, hoisting the rather length comments up to the top so as to not clutter the code so much whilst also rendering them more nicely in cheri-architecture.

CSeal_core no longer makes much sense now the length check (which happens quite late for CSeal) isn't shared.

### DIFF
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -699,7 +699,7 @@ function clause execute (CBuildCap(cd, cs1, cs2)) = {
     let (representable, cd2) = setCapOffset(cd1, getCapOffsetBits(cs2_val));
     let cd3 = setCapPerms(cd2, cs2_perms);
     let cd4 = setCapFlags(cd3, cs2_flags);
-    let cd5 = if cs2_val.otype == to_bits(cap_otype_width, otype_sentry) then sealCap(cd4, to_bits(cap_otype_width, otype_sentry)) else cd4;
+    let cd5 = if signed(cs2_val.otype) == otype_sentry then sealCap(cd4, to_bits(cap_otype_width, otype_sentry)) else cd4;
     {
       assert(exact, "CBuildCap: setCapBounds was not exact"); /* base and top came from cs2 originally so will be exact */
       C(cd) = cd5;
@@ -930,7 +930,7 @@ function clause execute (CCSeal(cd, cs1, cs2)) = {
   } else if cs2_cursor >= cs2_top then {
     C(cd) = cs1_val;
     RETIRE_SUCCESS
-  } else if cs2_val.address == to_bits(sizeof(xlen), otype_unsealed) then {
+  } else if signed(cs2_val.address) == otype_unsealed then {
     C(cd) = cs1_val;
     RETIRE_SUCCESS
   } else if isCapSealed(cs2_val) then {
@@ -1128,7 +1128,7 @@ function clause execute(CJALR(cd, cs1)) = {
   if not (cs1_val.tag) then {
     handle_cheri_reg_exception(CapEx_TagViolation, cs1);
     RETIRE_FAIL
-  } else if (isCapSealed(cs1_val) & cs1_val.otype != to_bits(cap_otype_width, otype_sentry)) then {
+  } else if isCapSealed(cs1_val) & (signed(cs1_val.otype) != otype_sentry) then {
     handle_cheri_reg_exception(CapEx_SealViolation, cs1);
     RETIRE_FAIL
   } else if not (cs1_val.permit_execute) then {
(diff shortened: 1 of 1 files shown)


src/cheri_insts.sail
```diff
@@ -745,70 +769,86 @@ function CSeal_core (cd, cs1, cs1_val, cs2, cs2_val) = {
...
-     * they wish to try, so just pass the result through.
-     */
+  } else if cs2_cursor < cs2_base then {
+    C(cd) = cs1_val;
+    RETIRE_SUCCESS
+  } else if cs2_cursor >= cs2_top then {
     C(cd) = cs1_val;
     RETIRE_SUCCESS
-  } else CSeal_core (cd, cs1, cs1_val, cs2, cs2_val)
+  } else if cs2_val.address == to_bits(sizeof(xlen), otype_unsealed) then {
```
PeterRugg, line 836: It would be nice if the source made it clearer whether this was 18 bits of 1s or xlen bits of 1s. Currently, I have to go and check whether otype_unsealed = -1 or 2^18-1

src/cheri_insts.sail
```diff
@@ -716,15 +716,39 @@ function clause execute (CSEQX(rd, cs1, cs2)) = {
...
+ * *cs2*.
+ *
+ * ## Exceptions
+ *
+ * An exception is raised if:
+ *   - *cs1*.**tag** is not set.
+ *   - *cs2*.**tag** is not set.
+ *   - *cs1*.**sealed** is set.
+ *   - *cs2*.**sealed** is set.
+ *   - *cs2*.**permit\_seal** is not set.
```
arichardson, line 732:
```suggestion
 *   - *cs2*.**permit_seal** is not set.
```
Do we have to escape the underscore? There aren't any more underscores in this line, so you'd hope the markdown parser doesn't do weird stuff.
It seems like sail uses a markdown parser that hasn't had a release since 2017 (https://opam.ocaml.org/packages/omd/#), but it seems like the recent git version claim to be commonmark 0.29 compliant. The spec says `Emphasis with _ is not allowed inside words`, so if this is a bug maybe we can upgrade the sail markdown parser to a git snapshot?

arichardson: I also wonder if we should be phrasing this more like `*cs* does not grant the Permit_Seal permission` in case we decide to merge redundant permission bits?

src/cheri_insts.sail
jrtc27: Well, all reserved otypes are sign-extended from bits(cap_otype_width) to xlenbits architecturally (e.g. in CGetLen) and defined as `2^XLEN - 1` etc in Software-Defined Permission Bits in 3.4.2. I guess as an alternative we could do:
```suggestion
  } else if signed(cs2_val.address) == otype_unsealed then {
```
if Sail permits that?

jrtc27: It compiles so I have switched to that as I do think it looks nicer. Do you find it clearer now?

PeterRugg: Yes, thank you! I actually thought it was the other way briefly. Was getting confused about needing to keep around all 2^xlen bits of bounds even though the typespace is only 18 bits, but of course in this case we don't care if the capability is in bounds anyway.

PeterRugg (approved):

src/cheri_insts.sail
```diff
@@ -745,70 +769,86 @@ function CSeal_core (cd, cs1, cs1_val, cs2, cs2_val) = {
...
-     */
+  let (cs2_base, cs2_top) = getCapBounds(cs2_val);
   if not (cs1_val.tag) then {
     handle_cheri_reg_exception(CapEx_TagViolation, cs1);
     RETIRE_FAIL
+  } else if not (cs2_val.tag) then {
+    /* CCopyType may not have been able to represent the result */
+    C(cd) = cs1_val;
+    RETIRE_SUCCESS
   } else if cs1_val.sealed then {
```
arichardson, line 827: Unrelated to this PR but maybe we should remove the separate sealed field and use a helper function instead? It made sense for 256 and 128 prior to the new format but I think now it's more likely to cause bugs then be useful.

arichardson (approved):
LGTM. We should probably make sure that the MIPS semantics match this so that the ISA spec is consistent.

src/cheri_insts.sail
jrtc27: Yeah it's a bug waiting to happen (and ISTR Peter did fall afoul of that recently?).

PeterRugg: Indeed, prompted the ugly "cap with otype=unsealed, sealed=false"

jrtc27: Done in f27e6397da

nwf (approved):

PR #34, opened 2020-10-04
Document more instructions

jrtc27:


src/cheri_insts.sail
```diff
@@ -917,6 +1063,23 @@ function clause execute (CInvoke(cs1, cs2)) = {
 }
 
 union clause ast = CJALR : (regidx, regidx)
+/*!
+ * Capability register *cd* is replaced with the next instruction's **PCC** and
+ * sealed as a sentry. **PCC** is replaced with the value of capability
+ * register *cs1* with the 0th bit of its **address** set to 0 and unsealed if
```
arichardson, line 1069:
```suggestion
 * register *cs1* with the 0th bit of its **address** set to 0 and is unsealed if
```

arichardson (commented):
LGTM, just a minor suggestion

arichardson (approved):

src/cheri_insts.sail
```diff
@@ -561,6 +606,20 @@ function clause execute (FPClear(q, m)) = {
 }
 
 union clause ast = CFromPtr : (regidx, regidx, regidx)
+/*!
+ * If the value of integer register *rs2* is 0 then capability register *cd* is
+ * set to **NULL**. Otherwise capability register *cd* is set to capability
+ * register *cs1* with its **offset** replaced with *rs2*. If the resulting
+ * capability cannot be represented exactly, then *cd*.**tag** is cleared. The
+ * remaining capability fields are set to what the in-memory representation of
+ * *cs1* with the address set to *cd*.**address** decodes to.
```
PeterRugg, line 615: It's almost a shame that the unrepresentable behaviour has to be described here, as it might confuse more casual readers a little. Maybe a paragraph break alone would help (the last sentence currently could apply in general or only to the unrepresentable case as it stands).

PeterRugg (approved):

src/cheri_insts.sail
jrtc27: Yeah, I just copied what's there for CIncOffset(Imm) and CSet{Offset,Addr}.

PR #35, opened 2020-10-05
Document the remaining ALU-ish instructions

jrtc27:
This leaves only the loads and stores undocumented.

### DIFF
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -152,6 +152,29 @@ function clause execute (CGetAddr(rd, cs1)) = {
 }
 
 union clause ast = CSpecialRW : (regidx, screg, regidx)
+/*!
+ * Capability register *cd* is set equal to special capability register *scr*,
+ * and *scr* is set equal to capability register *cs1* if *cs1* is not **C0**.
+ *
+ * ## Exceptions
+ *
+ * An exception is raised if:
+ *   - *scr* does not exist.
+ *   - *scr* is read-only and *cs1* is not **C0**.
+ *   - *scr* is only accessible to a higher privilege mode.
+ *   - *scr* requires Permit_Access_System_Registers and that is not granted by
+ *     **PCC**.**perms**.
+ *
+ * ## Notes
+ *
+ * Writing **NULL** to a special capability register cannot be done with **C0**
+ * as that only performs a read. An alternative implementation would allocate a
+ * separate two-operand CSpecialR instruction and interpret *cs1* being **C0**
+ * as a write of **NULL** if the need to use a temporary capability register
+ * proves to be overly problematic for software. For U-mode transitions to
+ * domains without Permit_Access_System_Registers only **DDC** should need
+ * clearing, which can be done with [CClear].
+ */
 function clause execute (CSpecialRW(cd, scr, cs1)) = {
   let (specialExists, ro, priv, needASR) : (bool, bool, Privilege, bool) = match unsigned(scr) {
     0  => (true, true,  User, false),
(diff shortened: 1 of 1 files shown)


src/cheri_insts.sail
```diff
@@ -152,6 +152,29 @@ function clause execute (CGetAddr(rd, cs1)) = {
 }
 
 union clause ast = CSpecialRW : (regidx, screg, regidx)
+/*!
+ * Capability register *cd* is set equal to special caapbility register *scr*,
```
arichardson, line 156:
```suggestion
 * Capability register *cd* is set equal to special capability register *scr*,
```

src/cheri_insts.sail
```diff
@@ -280,6 +303,26 @@ function clause execute(CSetFlags(cd, cs1, rs2)) = {
...
+ * An exception is raised if:
+ *   - *cs2*.**tag** is unset.
+ *   - *cs1*.**tag** is set and *cs1* is sealed.
+ *
+ * ## Notes
+ *
+ *   - *cs2* being sealed will not cause an exception to be raised. This is for
+ *     further study.
+ *
+ *   - [CGetAddr] similarly allows access to the sum of the base and offset of
```
arichardson, line 322: I think we can drop the base + offset bit? Also I don't really think this relates to cgetaddr, csub is much more similar.
Or maybe completely drop this MIPS note?

src/cheri_insts.sail
```diff
@@ -474,6 +547,20 @@ function clause execute (CSetBoundsImmediate(cd, cs1, uimm)) = {
 }
 
 union clause ast = CSetBoundsExact : (regidx, regidx, regidx)
+/*!
+ * Capability register *cd* is set to capability register *cs1* with its
+ * **base** field replaced with *cs1*.**address** and its **length** field
+ * replaced with integer register *rs2*.
```
arichardson, line 553: Maybe add something like this?

Unlike [csetbounds] this instruction will trap when the result cannot be represented exactly.

I know it's in the exceptions bit but I think most people will not bother reading that.

src/cheri_insts.sail
```diff
@@ -431,6 +474,21 @@ function clause execute (CSetAddr(cd, cs1, rs2)) = {
 
 
 union clause ast = CSetBounds : (regidx, regidx, regidx)
+/*!
+ * Capability register *cd* is set to capability register *cs1* with its
+ * **base** field replaced with *cs1*.**address** and its **length** field
+ * replaced with integer register *rs2*. If the resulting capability cannot be
+ * represented exactly the bounds are padded to be representable but will never
+ * exceed the bounds of *cs1*.
```
arichardson, line 482: The MIPS description has a bit more detail here. Should we add more on the rounding behaviour?
I think the current description is fine for now, but we should included more information when we update the spec to have MIPS as the appendix.

src/cheri_insts.sail
```diff
@@ -778,6 +865,20 @@ function clause execute(CRAM(rd, rs1)) = {
...
+ * Integer register *rd* is set to 1 if the **tag** fields of capability
+ * registers *cs1* and *cs2* are the same and the bounds and permissions of
+ * *cs2* are a subset of those of *cs1*.
+ *
+ * ## Notes
+ *
+ * - The operand order for this instruction is reversed compared with the
+ *   normal RISC-V comparison instructions, but this may be changed in future.
+ *
+ * - The **otype** field is ignored for this instruction, but an alternative
```
arichardson, line 878: That is rather surprising behaviour. I guess it is okay for sentries, but not necessarily for user defined otypes.

src/cheri_insts.sail
PeterRugg, line 322: address instead of "sum of the base and offset"?

jrtc27: Yeah I wasn't sure about it. I dropped a couple of others already, I'll just drop this too.

src/cheri_insts.sail
jrtc27: It's very difficult to describe precisely and accurately without just regurgitating all of CHERI Concentrate but I will add a bit more like CRRL and CRAM have.

src/cheri_insts.sail
jrtc27: Yeah, something I noticed whilst writing this so this note is an original creation rather than a regurgitation of something MIPS has.

arichardson (approved):

PeterRugg (approved):

PR #36, opened 2020-10-19
Legalise underivable caps

PeterRugg:
Tested to match the Bluespec cheri-cap-lib implementation (PR pending).

Also includes refactoring to allow an intermediate struct without any decoding. This is required to neatly clear the IE bit, but has the side-effect of reducing duplicated code in preludes.

### DIFF
--- a/src/cheri_cap_common.sail
+++ b/src/cheri_cap_common.sail
@@ -245,7 +245,7 @@ function getCapLegalised(c) : Capability -> Capability =
     c
   else
     /* Legalise the capability by zeroing the ie bit, implicitly setting E to 0. TODO avoid 90 magic number */
-    capBitsToCapability(false, capToBits(c) & ~(to_bits(128, 1) << 90))
+    encCapabilityToCapability(false, {capToEncCap(c) with internal_e = bitzero})
 
 function getCapBoundsBits(c) : Capability -> (CapAddrBits, CapLenBits) =
   let c_legal = getCapLegalised(c) in
--- a/src/cheri_cap_common.sail
+++ b/src/cheri_cap_common.sail
@@ -32,10 +32,94 @@
 /*  SUCH DAMAGE.                                                          */
 /*========================================================================*/
 
-
+let MAX_ADDR = MAX(cap_addr_width)
+type CapAddrInt = range(0, (2 ^ cap_addr_width) - 1)
+type CapLen = range(0, (2 ^ cap_len_width) - 1)
 let max_otype = MAX(otype_width) - reserved_otypes
 type CapAddrBits = bits(cap_addr_width)
 type CapLenBits  = bits(cap_len_width)
+
+/* A partially decompressed version of a capability -- E, B, T,
+ * lenMSB, sealed and otype fields are not present in all formats and are
+ * populated by capBitsToCapability.
+ */
+struct Capability = {
+  tag                    : bool    ,
+  uperms                 : bits(uperms_width) ,
+  permit_set_CID         : bool    ,
+  access_system_regs     : bool    ,
+  permit_unseal          : bool    ,
+  permit_cinvoke         : bool    ,
+  permit_seal            : bool    ,
+  permit_store_local_cap : bool    ,
+  permit_store_cap       : bool    ,
+  permit_load_cap        : bool    ,
+  permit_store           : bool    ,
+  permit_load            : bool    ,
+  permit_execute         : bool    ,
+  global                 : bool    ,
+  reserved               : bits(reserved_width) ,
+  flag_cap_mode          : bool    ,
+  internal_e             : bool    ,
+  E                      : bits(exp_width),
+  sealed                 : bool    ,
+  B                      : bits(mantissa_width),
+  T                      : bits(mantissa_width),
+  otype                  : bits(otype_width),
+  address                : bits(cap_addr_width)
+}
+
+let null_cap : Capability = struct {
+  tag                    = false,
+  uperms                 = zeros(),
+  permit_set_CID         = false,
+  access_system_regs     = false,
+  permit_unseal          = false,
+  permit_cinvoke         = false,
+  permit_seal            = false,
+  permit_store_local_cap = false,
+  permit_store_cap       = false,
+  permit_load_cap        = false,
+  permit_store           = false,
+  permit_load            = false,
+  permit_execute         = false,
+  global                 = false,
+  reserved               = zeros(),
+  flag_cap_mode          = false,
+  internal_e             = true,
+  E                      = resetE,
+  sealed                 = false,
+  B                      = zeros(),
+  T                      = resetT,
+  otype                  = to_bits(otype_width, otype_unsealed),
+  address                = zeros()
+}
+
+let default_cap : Capability = struct {
+  tag                    = true,
+  uperms                 = ones(),
+  permit_set_CID         = true,
+  access_system_regs     = true,
+  permit_unseal          = true,
+  permit_cinvoke         = true,
+  permit_seal            = true,
+  permit_store_local_cap = true,
+  permit_store_cap       = true,
+  permit_load_cap        = true,
+  permit_store           = true,
+  permit_load            = true,
+  permit_execute         = true,
+  global                 = true,
+  reserved               = zeros(),
+  flag_cap_mode          = false,
+  internal_e             = true,
+  E                      = resetE,
+  sealed                 = false,
+  B                      = zeros(),
+  T                      = resetT,
+  otype                  = to_bits(otype_width, otype_unsealed),
+  address                = zeros()
+}
 
 function getCapHardPerms(cap) : Capability -> bits(12) =
    (bool_to_bits(cap.permit_set_CID)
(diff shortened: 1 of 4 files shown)


src/cheri_cap_common.sail
```diff
@@ -88,41 +228,48 @@ function capToMemBits(cap) : Capability -> CapBits =
...
+    (base, top)
+
+function getCapLegalised(c) : Capability -> Capability =
+  let (base, top) = getCapBoundsBitsUncorrected(c) in
+  /* Check that base is within the address space */
+  if unsigned(c.E) <= maxE
+   & (base <_u (to_bits(cap_len_width, 1) << cap_addr_width)) then
+    c
+  else
+    /* Legalise the capability by zeroing the ie bit, implicitly setting E to 0. TODO avoid 90 magic number */
```
arichardson, line 247: I don't see a 90 magic number?

arichardson:
I'm not sure how useful it is to share struct Capability between 64 and 128. We might provide a subset of permissions for 64 in the future? But I guess we could always hardcode the others to false.

Also this PR needs to be rebased.

src/cheri_cap_common.sail
jrtc27: Maybe that's why it's magic :P but I assume this used to do a `1 << 90` to set IE in the compressed 128-bit representation.

PeterRugg: Yep, good spot. I'll remove the TODO

PeterRugg:
> I'm not sure how useful it is to share struct Capability between 64 and 128. We might provide a subset of permissions for 64 in the future? But I guess we could always hardcode the others to false.
> 
> Also this PR needs to be rebased.

Hmm, yes, or factor out the permissions into a separate struct in that case. It seemed silly to duplicate all the decoding logic at least...

I'll do a rebase now. It seems to mostly be trivial mantissa_width vs cap_mantissa_width.

src/cheri_cap_common.sail
```diff
@@ -32,21 +32,105 @@
 /*  SUCH DAMAGE.                                                          */
 /*========================================================================*/
 
-
+let MAX_ADDR = MAX(cap_addr_width)
```
arichardson, line 35: We should use consistent naming. In this file we have 3 conflicting styles now: `MAX_ADDR`, `max_otype` and `maxE`.

PeterRugg: Agreed!

PeterRugg: Do you think the cap_ prefix should be used for these also?

arichardson: That probably makes sense. It also makes it less likely to conflict with any future base riscv extensions.

src/cheri_cap_common.sail
```diff
@@ -61,30 +145,85 @@ function getCapHardPerms(cap) : Capability -> bits(cap_hperms_width) =
...
-  t_hi : bits(cap_mantissa_width - 5) = cap.T[cap_mantissa_width - 3..3];
-  t_lo : bits(3)  = cap.T[2..0];
-  b_hi : bits(cap_mantissa_width - 3) = cap.B[cap_mantissa_width - 1..3];
-  b_lo : bits(3)  = cap.B[2..0];
-  if cap.internal_e then {
-    t_lo = cap.E[5..3];
-    b_lo = cap.E[2..0];
+/* Convert from capabilty bits (128 bits with tag) to a more convenient struct. */
+function encCapabilityToCapability(t,c) : (bool, EncCapability) -> Capability = {
+  let sealed : bool = c.otype != to_bits(cap_otype_width, otype_unsealed);
```
arichardson, line 150: sealed seems unused?

src/cheri_cap_common.sail
```diff
@@ -61,30 +145,85 @@ function getCapHardPerms(cap) : Capability -> bits(cap_hperms_width) =
...
+    lenMSBs = 0b00;
+    T = c.T;
+    Bs = c.B;
+  };
+  /* Reconstruct top two bits of T given T = B + len and:
+   * 1) the top two bits of B
+   * 2) most significant two bits of length derived from format above
+   * 3) carry out of B[20..0] + len[20..0] that is implied if T[20..0] < B[20..0]
+   */
+  carry_out = if T <_u Bs[cap_mantissa_width - 3..0] then 0b01 else 0b00;
```
arichardson, line 173:
```suggestion
  carry_out = if T <_u Bs[cap_mantissa_width - internal_E_take_bits..0] then 0b01 else 0b00;
```

src/cheri_cap_common.sail
```diff
@@ -61,30 +145,85 @@ function getCapHardPerms(cap) : Capability -> bits(cap_hperms_width) =
...
+  Ttop2 = Bs[cap_mantissa_width - 1..cap_mantissa_width - 2] + lenMSBs + carry_out;
```
arichardson, line 174:
```suggestion
  Ttop2 = Bs[cap_mantissa_width - 1..cap_mantissa_width - internal_E_take_bits - 1] + lenMSBs + carry_out;
```

src/cheri_cap_common.sail
PeterRugg: No, I think these are unrelated. The top three bits are used to divide the representable region into eighths. While that could be parametrisable, it's independent of the number of bits used in the internal exponent I believe?

src/cheri_cap_common.sail
PeterRugg: Yep, I missed the change in the rebase.

src/cheri_cap_common.sail
arichardson: I just saw 3 and assumed it was related.

bauereiss (approved):
Looks good to me, I've just left a few syntactic suggestions.  The properties I've been SMT-checking before still hold, and I was able to strengthen the "(base <= top)" property by removing the precondition that the capability has canonical bounds, which is good.

src/cheri_cap_common.sail
```diff
@@ -32,21 +32,105 @@
...
+let cap_max_E   = cap_len_width - cap_mantissa_width
+let cap_reset_E = to_bits(cap_E_width, cap_max_E)
+let cap_reset_T = 0b01 @ zeros(cap_mantissa_width - 2)
+
+/* A partially decompressed version of a capability -- E, B, T,
+ * lenMSB, sealed and otype fields are not present in all formats and are
+ * populated by capBitsToCapability.
+ */
+struct Capability = {
+  tag                    : bool    ,
```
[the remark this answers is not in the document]
PeterRugg: I think I inherited some weirdness, but I agree, it doesn't make sense. I'll push something cleaner.

src/cheri_cap_common.sail
```diff
@@ -61,30 +145,84 @@ function getCapHardPerms(cap) : Capability -> bits(cap_hperms_width) =
...
-    @ t_hi
-    @ t_lo
-    @ b_hi
-    @ b_lo
-    @ cap.address
-  );
 }
 
+/* Convert from capabilty struct to bits (no tag) */
+function capToBits(cap) : Capability -> CapBits = EncCapToBits(capToEncCap(cap))
```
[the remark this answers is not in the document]
PeterRugg: Agreed!

src/cheri_cap_common.sail
```diff
@@ -61,30 +145,84 @@ function getCapHardPerms(cap) : Capability -> bits(cap_hperms_width) =
...
-/* Convert from capabilty struct to bits (no tag) */
-function capToBits(cap) : Capability -> CapBits = {
-  t_hi : bits(cap_mantissa_width - 5) = cap.T[cap_mantissa_width - 3..3];
-  t_lo : bits(3)  = cap.T[2..0];
-  b_hi : bits(cap_mantissa_width - 3) = cap.B[cap_mantissa_width - 1..3];
-  b_lo : bits(3)  = cap.B[2..0];
-  if cap.internal_e then {
-    t_lo = cap.E[5..3];
-    b_lo = cap.E[2..0];
+/* Convert from capabilty bits (128 bits with tag) to a more convenient struct. */
```
arichardson, line 148:
```suggestion
/* Convert from capability bits (128 bits with tag) to a more convenient struct. */
```

src/cheri_cap_common.sail
```diff
@@ -61,30 +145,84 @@ function getCapHardPerms(cap) : Capability -> bits(cap_hperms_width) =
...
-    @ bool_to_bits(cap.internal_e)
-    @ t_hi
-    @ t_lo
-    @ b_hi
-    @ b_lo
-    @ cap.address
-  );
 }
 
+/* Convert from capabilty struct to bits (no tag) */
```
arichardson, line 223:
```suggestion
/* Convert from capability struct to bits (no tag) */
```

arichardson (commented):
Seems fine to me, I'll re-generate the sail->C and fuzz against cheri-compressed-cap to see how long it takes libfuzzer to find a mismatch.

PeterRugg:
> Seems fine to me, I'll re-generate the sail->C and fuzz against cheri-compressed-cap to see how long it takes libfuzzer to find a mismatch.

Sounds good, although we do expect differences: it might take some effort to differentiate bugs from intended changes. I haven't looked at the cheri-compressed-cap code much, but could add the changes to that if needed.

PeterRugg:
@jrtc27 It would be nice to get this merged before too much rebasing is required. Any chance you're going to have time to take a look at it in the near future?

PeterRugg:
I've tidied up the history and pulled most of the uncontroversial contents into master.

src/cheri_cap_common.sail
```diff
@@ -236,41 +236,48 @@ function capToMemBits(cap) : Capability -> CapBits =
...
-       to the one above, but differs at least in some cases where E >= cap_max_E - 1
-    if (base[cap_addr_width] == bitone) then {
-       top[cap_addr_width] = if (aHi == 1) & (tHi == 1) then bitone else bitzero;
-    };
-    */
-    (base[(cap_addr_width - 1)..0], top)
+  let correction_top  = tHi - aHi in {
+    base_corrected : CapLenBits = base + (to_bits(cap_len_width, correction_base) << (cap_mantissa_width + E));
+    top_corrected : CapLenBits = top + (to_bits(cap_len_width, correction_top) << (cap_mantissa_width + E));
+    /* Interpret the top relative to the address space of the base.
```
arichardson, line 277: Could you expand on this comment, e.g. explaining why this prevents base > top?

src/cheri_cap_common.sail
```diff
@@ -236,41 +236,48 @@ function capToMemBits(cap) : Capability -> CapBits =
...
+       This corrects for errors that happen when the representable space wraps the address space. */
+    top_corrected[cap_len_width - 1 .. cap_addr_width] = top_corrected[cap_len_width - 1 .. cap_addr_width] - base_corrected[cap_len_width - 1 .. cap_addr_width];
```
arichardson, line 279: This is only one bit, I don't think it could it ever be more than one? Maybe it would be simpler to use `top_corrected[cap_addr_width] = ...` (assuming we have an assertion that len_width == addr_width + 1 somewhere else)?

PeterRugg: I believe it's 2 bits (https://github.com/CTSRD-CHERI/sail-cheri-riscv/blob/f3a264c1966fcdc3005f889abdd40ff293b89ab8/src/cheri_prelude_128.sail#L58)? For cap64, cap_len_width is 66, so this is bits [65..64].

src/cheri_cap_common.sail
PeterRugg: Hmm, I don't think this alone prevents that. It's more ensuring that the top hasn't wrapped around to be in an entire address space (or more) below the base? I must admit, my memory is slightly hazy.

arichardson: Yeah it's been a while since I looked at this which is why I wasn't sure why we can now assert that base <= top. Adding some comments explaining where this is guaranteed would be helpful for future readers I believe.

src/cheri_cap_common.sail
arichardson: Oh yeah we want two bits at the top. Please ignore...

src/cheri_cap_common.sail
PeterRugg: Right, agreed. I guess "it's been proved" is something, but it's also nice to see how it's been constructed to make this true.

PR #37, opened 2020-10-23
Add saildoc for non-atomic loads and stores

jrtc27:
I have deliberately omitted any details that are not specific to CHERI (i.e. I don't document alignment requirements, translation exceptions or load/store faults as they all apply to things like the normal LW and SW, except for highlighting that capability loads/stores must always be aligned).

### DIFF
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -1305,6 +1305,20 @@ function handle_load_data_via_cap(rd, auth_idx, auth_val, vaddrBits, is_unsigned
 }
 
 union clause ast = LoadDataDDC : (regidx, regidx, bool, word_width)
+/*!
+ * Integer register *rd* is replaced with the signed or unsigned byte,
+ * halfword, word or doubleword located in memory at **DDC**.**address** $+$
+ * *rs1*.
+ *
+ * ## Exceptions
+ *
+ * An exception is raised if:
+ *   - **DDC**.**tag** is not set.
+ *   - **DDC** is sealed.
+ *   - **DDC**.**perms** does not grant **Permit_Load**.
+ *   - **DDC**.**address** $+$ *rs1* $\lt$ **DDC**.**base**.
+ *   - **DDC**.**address** $+$ *rs1* $+$ *size* $\gt$ **DDC**.**top**.
+ */
 function clause execute (LoadDataDDC(rd, rs1, is_unsigned, width)) = {
   let ddc_val = DDC;
   let vaddr = ddc_val.address + X(rs1);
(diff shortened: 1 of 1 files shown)


jrtc27:
Currently working through the others but figured I should publish this now to pipeline the review process.

src/cheri_insts.sail
```diff
@@ -1305,13 +1305,40 @@ function handle_load_data_via_cap(rd, auth_idx, auth_val, vaddrBits, is_unsigned
 }
 
 union clause ast = LoadDataDDC : (regidx, regidx, bool, word_width)
+/*!
+ * Integer register *rd* is replaced with the signed or unsigned byte,
```
PeterRugg, line 1309: It seems redundant to list all this in detail here when it's just a call through to LoadDataCap, which already has the description. By no means the end of the world though.

src/cheri_insts.sail
```diff
@@ -1357,13 +1384,45 @@ function handle_load_cap_via_cap(cd, auth_idx, auth_val, vaddrBits) = {
 }
 
 union clause ast = LoadCapDDC : (regidx, regidx)
+/*!
+ * Capability register *cd* is replaced with the capability located in memory
```
PeterRugg, line 1388: Same redundancy as above.

arichardson:
LGTM. I do wonder if we want to create a link for Permit_Load, etc.

PeterRugg (approved):

jrtc27:
I have magic to make them use the `\cappermG` etc macros (well, aliases for them) so they get formatted identically as elsewhere in the whole spec.

src/cheri_insts.sail
jrtc27: It calls the helper, not LoadDataCap. Those helpers don't have any saildoc, nor is any additional documentation given to them in cheri-architecture. Also they appear in an earlier section, whereas this description goes on the actual instruction specification page(s), so I think there is value in detailing everything there even if the handle functions were also documented.

PeterRugg: Oh yes, sorry, misread. It still seems like ideally these would be refactored to avoid the redundancy, but I can't see directly an elegant way to achieve that.

jrtc27: Yeah I think that's too much magic to get it to work with cheri-architecture's layout. Especially given we also have the mode-dependent instructions.

jrtc27:
Now with everything. Lots of copy-paste involved so please keep an eye out for related errors :)

src/cheri_insts.sail
```diff
@@ -1639,13 +1758,74 @@ function clause execute AUIPCC(imm, cd) = {
...
+ * *cd*.**tag** is cleared.
+ *
+ * ## Exceptions
+ *
+ * In integer mode, an exception is raised if:
+ *   - **DDC**.**tag** is not set.
+ *   - **DDC** is sealed.
+ *   - **DDC**.**perms** does not grant **Permit_Load**.
+ *   - **DDC**.**address** $+$ *rs1* $+$ *imm* $\lt$ **DDC**.**base**.
+ *   - **DDC**.**address** $+$ *rs1* $+$ *imm* $+$ **CLEN** $\gt$
```
PeterRugg, line 1777: Is it obvious that CLEN is in bytes here where normally it's bits? Are we okay for context to make this clear or should we have a disambiguation (e.g. XLEN vs xlen)?

jrtc27: Oh that's a good point...

jrtc27: I could do `CLEN / 8`?

PeterRugg (approved):

src/cheri_insts.sail
PeterRugg: Hmm, I suppose. CLEN / BITS_PER_BYTE seems overly verbose...

jrtc27: Not particularly keen to encourage things akin to C's `CHAR_BITS`... :joy:

PR #38, opened 2020-10-26
CLoadTags

nwf:


src/cheri_prelude_128.sail
```diff
@@ -52,6 +52,10 @@ let  cap_len_width  = sizeof(cap_len_width)
 type cap_mantissa_width : Int = 14
 let  cap_mantissa_width = sizeof(cap_mantissa_width)
 
+/* For CLoadTags and CClearTags, which operate at cacheline granularity */
+type caps_per_cacheline : Int = 8
```
jrtc27, line 56: All our implementations have 512b cache lines, so 4 caps per line on CHERI-128.

src/cheri_insts.sail
```diff
@@ -1370,6 +1370,66 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+    handle_cheri_reg_exception(CapEx_SealViolation, cs1);
+    RETIRE_FAIL
+  } else if not (cs1_val.permit_load) then {
+    handle_cheri_reg_exception(CapEx_PermitLoadViolation, cs1);
+    RETIRE_FAIL
+  } else if not (cs1_val.permit_load_cap) then {
+    handle_cheri_reg_exception(CapEx_PermitLoadCapViolation, cs1);
+    RETIRE_FAIL
+  } else {
+    let vaddr = cs1_val.address;
```
jrtc27, line 1389: This can be hoisted to avoid indenting (as is effectively done with the normal loads/stores due to calling the helpers with the desired vaddr).

src/cheri_insts.sail
```diff
@@ -1370,6 +1370,66 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+    if not (inCapBounds(cs1_val, vaddr, caps_per_cacheline * cap_size)) then {
+      handle_cheri_reg_exception(CapEx_LengthViolation, cs1);
+      RETIRE_FAIL
+    } else if not (unsigned(vaddr) % (caps_per_cacheline * cap_size) == 0) then {
+      handle_mem_exception(vaddr, E_Load_Addr_Align());
+      RETIRE_FAIL
+    } else match translateAddr(vaddr, Read(Cap)) {
+      TR_Failure(e, _) => { handle_mem_exception(vaddr, e); RETIRE_FAIL },
```
jrtc27, line 1397: For handle_load_cap_via_cap we have a case for E_extension that gives an internal_error; should be replicated here.

src/cheri_insts.sail
```diff
@@ -1370,6 +1370,66 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+      TR_Address(addr, ptw_info) => {
+	match ptw_info.ptw_lc {
+          PTW_LC_CLEAR => {
+            handle_mem_exception(vaddr, E_Extension(EXC_LOAD_CAP_PAGE_FAULT));
+            RETIRE_FAIL
+          },
+          PTW_LC_TRAP => {
+            handle_mem_exception(vaddr, E_Extension(EXC_LOAD_CAP_PAGE_FAULT));
+            RETIRE_FAIL
+          },
```
jrtc27, line 1407: `if ptw_info.ptw_lc != PTW_LC_OK`?

src/cheri_insts.sail
```diff
@@ -1370,6 +1370,66 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+            foreach (i from 0 to (caps_per_cacheline - 1)) {
+              let c = mem_read_cap(addr, false, false, false);
+              match (x,c) {
+		(MemException(e), _) => { () },
+                (MemValue(xv), MemValue(cv)) => {
+                  y = xv; y[i] = bool_to_bits(cv.tag)[0]; x = MemValue(y)
+                },
+                (_, MemException(e)) => { x = MemException(e) }
+              }
+            };
```
jrtc27, line 1420: Do we really not have bind or an imperative equivalent? We should have one and remove this gunk... I guess in this case what we really want is a `sequence :: Monad m => [m a] -> m [a]` equivalent; could put one in cheri_mem.sail and then eventually move it into riscv_mem.sail. Then you could write:
```sail
res : vector(caps_per_cacheline, dec, MemoryOpResult(bit));
foreach (i from 0 to caps_per_cacheline)
  res[i] = mem_read_cap(addr, false, false, false);

match MemoryOpResult_sequence(res) {
  ...
}
```

I'll have a play with that :)

src/cheri_insts.sail
```diff
@@ -1370,6 +1370,66 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+          PTW_LC_CLEAR => {
+            handle_mem_exception(vaddr, E_Extension(EXC_LOAD_CAP_PAGE_FAULT));
+            RETIRE_FAIL
+          },
+          PTW_LC_TRAP => {
+            handle_mem_exception(vaddr, E_Extension(EXC_LOAD_CAP_PAGE_FAULT));
+            RETIRE_FAIL
+          },
+          PTW_LC_OK => {
+            x : MemoryOpResult(xlenbits) = MemValue(zeros());
```
jrtc27, line 1409: I'd prefer to make this `caps_per_cacheline` and zero-extend at the end.

src/cheri_insts.sail
```diff
@@ -1863,6 +1923,7 @@ mapping clause encdec = CMove(cd, cs1)      if (haveXcheri()) <-> 0b1111111 @ 0b
 mapping clause encdec = CClearTag(cd, cs1)  if (haveXcheri()) <-> 0b1111111 @ 0b01011 @ cs1 @ 0b000 @ cd @ 0b1011011 if (haveXcheri())
 mapping clause encdec = CJALR(cd, cs1)      if (haveXcheri()) <-> 0b1111111 @ 0b01100 @ cs1 @ 0b000 @ cd @ 0b1011011 if (haveXcheri())
 mapping clause encdec = CSealEntry(cd, cs1) if (haveXcheri()) <-> 0b1111111 @ 0b10001 @ cs1 @ 0b000 @ cd @ 0b1011011 if (haveXcheri())
+mapping clause encdec = CLoadTags(rd, cs1)  if (haveXcheri()) <-> 0b1111111 @ 0b10010 @ cs1 @ 0b000 @ rd @ 0b1011011 if (haveXcheri())
```
jrtc27, line 1926: I'd separate it given it's rd not cd and very unrelated (and then drop the extra space).

src/cheri_insts.sail
```diff
@@ -1886,6 +1947,7 @@ mapping clause assembly = CClearTag(cd, cs1)  <-> "ccleartag"  ^ spc() ^ cap_reg
 mapping clause assembly = CJALR(0b00000, cs1) <-> "cjr"        ^ spc() ^ cap_reg_name(cs1)
 mapping clause assembly = CJALR(cd, cs1)      <-> "cjalr"      ^ spc() ^ cap_reg_name(cd) ^ sep() ^ cap_reg_name(cs1)
 mapping clause assembly = CSealEntry(cd, cs1) <-> "csealentry" ^ spc() ^ cap_reg_name(cd) ^ sep() ^ cap_reg_name(cs1)
+mapping clause assembly = CLoadTags(rd, cs1)  <-> "cloadtags"  ^ spc() ^ reg_name(rd) ^ sep() ^ cap_reg_name(cs1)
```
jrtc27, line 1950: Ditto.

jrtc27:
This will screw up RVFI but that's a problem for another day (and only matters once QCVEngine starts generating them).

src/cheri_prelude_128.sail
nwf: I was about to say "there's no particular reason that Sail has to match any particular implementation" but I guess we want that for TestRIG.  Maybe the answer is that there should be a WARL csr field somewhere and sail can actually use that value while hardware hard-codes it.

src/cheri_insts.sail
nwf: I confess my familiarity with sail's base library is... lacking.  If there's a `sequence` laying around, I'm not aware of it.  If there's also a `bind`, we could simplify some of the PTW logic the Revocation 3 PTE bits have introduced.

src/cheri_prelude_128.sail
jrtc27: It's something for riscv-config when that materialises any year now, just like all the other bits of implementation-defined behaviour.

src/cheri_insts.sail
jrtc27: (sequence didn't make it much better)

jrtc27:
So there's still the... interesting point of what happens if you CLoadTags from I/O memory. Sail will let you do that and treat it like a series of reads, but an optimised hardware implementation wouldn't pull in the data and just go to the tag controller which would return 0s due to not being covered. I don't think that's a blocker for merging, but something to ponder... ideally we'd just fault because it's a stupid thing to do.

src/cheri_insts.sail
```diff
@@ -1430,6 +1430,84 @@ function clause execute (LoadCapCap(cd, cs1)) = {
   handle_load_cap_via_cap(cd, 0b0 @ cs1, cs1_val, vaddr)
 }
 
+union clause ast = CLoadTags : (regidx, regidx)
+/*!
+ * Integer register *rd* is replaced with the tags of the capabilities located
+ * at and above *cs1*.**address**.  The 0th bit corresponds to the first
```
jrtc27, line 1436:
```suggestion
 * in memory at and above *cs1*.**address**. The 0th bit corresponds to the first
```

src/cheri_insts.sail
```diff
@@ -1430,6 +1430,84 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+ * capability in memory.  The result is coherent with other processors, *as if*
```
jrtc27, line 1437:
```suggestion
 * capability in memory. The result is coherent with other processors, as if
```
not keen on how MIPS styles it.

src/cheri_insts.sail
```diff
@@ -1430,6 +1430,84 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+ * the corresponding data words had also been loaded.  The number of tags
```
jrtc27, line 1438:
```suggestion
 * the corresponding data words had also been loaded. The number of tags
```

src/cheri_insts.sail
```diff
@@ -1430,6 +1430,84 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+ * loaded is implementation defined; typical implementations are expected to
```
jrtc27, line 1439:
```suggestion
 * loaded is implementation-defined; typical implementations are expected to
```

src/cheri_insts.sail
```diff
@@ -1430,6 +1430,84 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+ * return the tags held in a L1 cache line, and so we use the sail constant
```
jrtc27, line 1440:
```suggestion
 * return the tags held in an L1 cache line, and so we use the constant
```
(sail is the language)

src/cheri_insts.sail
```diff
@@ -1430,6 +1430,84 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+ * *caps_per_cacheline*.  The number of tags loaded must be a power of two, at
```
jrtc27, line 1441:
```suggestion
 * [caps_per_cacheline]. The number of tags loaded must be a power of two, at
```

src/cheri_insts.sail
```diff
@@ -1430,6 +1430,84 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+ * least 1, and no more than **XLEN**.
+ *
+ * ## Exceptions
+ *
+ * An exception is raised if:
+ *   - *cs1*.**tag** is not set.
+ *   - *cs1* is sealed.
+ *   - *cs1* does not grant both **Permit_Load** and
```
jrtc27, line 1449:
```suggestion
 *   - *cs1*.**perms** does not grant both **Permit_Load** and
```

src/cheri_insts.sail
```diff
@@ -1430,6 +1430,84 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+ *     **Permit_Load_Capabilities**.
+ *   - *cs1*.**address** $\lt$ *cs1*.**base**.
+ *   - *cs1*.**address** $+$ **caps_per_cacheline* $*$ **CLEN** $/$ 8 $\gt$ *cs1*.**top**.
+ *   - *cs1*.**address** is not properly aligned.
```
jrtc27, line 1453:
```suggestion
 *   - *cs1*.**address** is unaligned.
```

src/cheri_insts.sail
```diff
@@ -1430,6 +1430,84 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+ *   - The page mapping for *cs1*.**address** specifies that capability loads are to
+ *     to trigger traps or clear tags.
```
jrtc27, line 1455:
```suggestion
 *   - The page table entry for *cs1*.**address** would cause the tag to be cleared.
```
since the other traps are normal for loads (and we don't document the MMU faults in every load/store instruction, since they come from the MMU)

src/cheri_insts.sail
```diff
@@ -1430,6 +1430,83 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+
+        foreach (i from 0 to (caps_per_cacheline - 1)) {
+          match mtags {
+            MemException(_) => (),
+            MemValue(tags) => {
+              match mem_read_cap(addr + i * cap_size, aq, rl, false) {
+                MemException(e) => mtags = MemException(e),
+                MemValue(v) => mtags = MemValue([tags with i..i = bool_to_bits(v.tag)])
+              }
+            }
```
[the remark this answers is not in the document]
jrtc27: Yeah... it's no word than handle_store_cond_cap_via_cap...

src/cheri_insts.sail
```diff
@@ -1430,6 +1430,110 @@ function clause execute (LoadCapCap(cd, cs1)) = {
...
+ *     $\gt$ *cs1*.**top**.
+ *   - *cs1*.**address** is unaligned.
+ *   - The page table entry for *cs1*.**address** would cause the tag to be
+ *     cleared.
+ *
+ * ## Notes
+ *
+ * - In order to reduce DRAM traffic, implementations may choose to load only
+ *   the tags and not the corresponding data, and may wish to not evict other
+ *   lines by treating it as either an uncached or a non-temporal/streaming
```
nwf, line 1466: "uncached" apparently means "skipping over any cached values" in some contexts and so might be confusing.  Maybe say "non-evicting" or something?

nwf (commented):
One nit; otherwise LGTM.

src/cheri_insts.sail
jrtc27: I think I'll just drop the uncached part, non-temporal/streaming can encompass that behaviour.

PR #39, opened 2020-11-14
CBuildCap: Raise an exception if reserved bits are set

arichardson:
This raises a LengthViolation for now, but we may want to allocate a new
exception code for reserved bits.


PeterRugg:
This is already solved by https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/36, which also solves the issue for other non-canonical capabilities

arichardson:
Indeed, deriving from cs1 instead of setting the tag bit (which is what QEMU does) fixes that problem. Should probably change qemu to re-derive as well, that way we also get the canonicalization.

PR #40, opened 2020-11-14
Report CHERI exceptions in xTVAL

arichardson:
The matches the latest ISA spec and QEMU

### DIFF
--- a/src/cheri_addr_checks.sail
+++ b/src/cheri_addr_checks.sail
@@ -10,7 +10,7 @@ function handle_cheri_cap_exception(capEx, regnum) =
     };
     let t : sync_exception = struct {
       trap    = E_Extension(EXC_CHERI),
-      excinfo = (None() : option(xlenbits)),
+      excinfo = Some(EXTZ(regnum @ CapExCode(capEx)) : xlenbits),
       ext     = Some(cause)
     };
     set_next_pc(exception_handler(cur_privilege, CTL_TRAP(t), PC))


PeterRugg:
Hmm, does this duplicate the info between xccsr and xtval? I would expect we would remove the reporting from xccsr alongside adding it to xtval?

arichardson:
We currently still have both in the ISA spec, but I agree it's redundant.

nwf:
Can we get rid of the xccsr bits from the spec?  Having to update just one CSR on exception is, I think, more "the RISC-V way" and would simplify the work I need to do for #18. ;)

arichardson:
> Can we get rid of the xccsr bits from the spec? Having to update just one CSR on exception is, I think, more "the RISC-V way" and would simplify the work I need to do for #18. ;)

Yes that would be great. I found a bug in qemu that I won't have to fix if we do that :)

bsdjhb:
I would be happy if we removed the xccsr bits.  I had deprecated them in ISAv8 to give us time to make the transition in all our implementations.  I suspect our P1 and CheriFreeRTOS still use xccsr?

jrtc27:
> I would be happy if we removed the xccsr bits. I had deprecated them in ISAv8 to give us time to make the transition in all our implementations. I suspect our P1 and CheriFreeRTOS still use xccsr?

It seems Piccolo hasn't pulled that change in even in the source (yay don't you love code copies). CheriFreeRTOS only looks at the CHERI cause if you use the simple sandbox mechanism so yes it will break but nothing actually uses that other than the old secure ballot box system.

arichardson:
@jrtc27 Is this okay to merge now? I'll open a separate PR for the xccsr changes.

jrtc27 (approved):

PR #42, opened 2020-11-24
Stop reporting cap_cause and cap_index in xCCSR

arichardson:
These values are reported in xtval, so this information is redundant.


src/cheri_sys_exceptions.sail
```diff
@@ -9,31 +9,16 @@ type ext_exception = cheri_cause
 function handle_trap_extension(p : Privilege, pc : xlenbits, ccause : option(cheri_cause)) -> unit = {
```
jrtc27, line 9: Can we now axe cheri_cause entirely and replace it with `unit` like in the base RISC-V model?

jrtc27 (approved):

PR #43, opened 2020-11-24
Don't crash when decoding an invalid cspecialrw

arichardson:
It appears that using a wildcard match triggers a sail C code generation
bug that results in and invalid call to hex_bits_5_matches_prefix()
being generated.

Enumerating all options is an awful workaround but better than crashing.


rmn30:
This looks like a suitable workaround. Is this the sail bug: https://github.com/rems-project/sail/issues/62 ?

arichardson:
That does sound like the same bug.

arichardson:
@jrtc27 Okay to merge or should we wait for a sail bugfix?

jrtc27:
Yeah this seems fine, it's a long-standing Sail bug that won't be fixed any time soon (and is actually rather difficult to fix properly IIRC...).

PR #45, opened 2021-05-04
Add mode-dependent jumps

jrtc27:


nwf (approved):
LGTM

rmn30:
Looks fine. Can't tell if there were any changes to the code that moved (CJALR and AUIPCC executes).

src/cheri_insts.sail
```diff
@@ -32,6 +32,108 @@
...
+ *   - *cs1*.**tag** is not set.
+ *   - *cs1* is sealed and is not a sentry.
+ *   - *cs1* is a sentry and *imm* $\ne$ 0.
+ *   - *cs1*.**perms** does not grant **Permit_Execute**.
+ *   - *cs1*.**address** $+$ *imm* $\lt$ *cs1*.**base**.
+ *   - *cs1*.**address** $+$ *imm* $+$ min_instruction_bytes $\gt$ *cs1*.**top**.
+ *   - *cs1*.**base** is unaligned.
+ *   - *cs1*.**address** $+$ *imm* is unaligned, ignoring bit 0.
+ */
+function clause execute(CJALR(imm, cs1, cd)) = {
```
arichardson, line 102: I would have expected the arguments to be `imm, cd, cs1` or `cd, cs1, imm` but it matches the base instructions so I guess that's consistent.

arichardson (approved):

src/cheri_insts.sail
jrtc27: Yeah... sail-riscv puts them in encoding order, we put them generally in assembly order. For AUIPCC we followed AUIPC, so I did the same for CJAL[R]. We might want to reconsider that choice for them though.

PR #47, opened 2021-05-20
Add cap-mode dependent lr / sc and AMOSWAP

### LINKED ISSUE #46: Support for CHERI AMOs
The model needs to support capability AMO isntructions as specified in section C.2.2 of the cheri architecture manual.

rmn30:
Partially fixes #46 . It is only partial because `{lr, sc}.{b, h}` still have to be implemented. The will require a tweak to the sail-riscv spec.

### DIFF
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -1671,10 +1671,8 @@ function handle_loadres_data_via_cap(rd, auth_idx, auth_val, vaddrBits, width) =
   }
 }
 
-val handle_loadres_cap_via_cap : (regidx, capreg_idx, Capability, xlenbits) -> Retired effect {escape, rmem, rmemt, rreg, wmv, wmvt, wreg}
-function handle_loadres_cap_via_cap(cd, auth_idx, auth_val, vaddrBits) = {
-  let aq : bool = true;  /* cheri-specific aq/rl */
-  let rl : bool = true;
+val handle_loadres_cap_via_cap : (regidx, capreg_idx, Capability, xlenbits, bool, bool) -> Retired effect {escape, rmem, rmemt, rreg, wmv, wmvt, wreg}
+function handle_loadres_cap_via_cap(cd, auth_idx, auth_val, vaddrBits, aq, rl) = {
   let is_unsigned = false;
 if not(auth_val.tag) then {
     handle_cheri_cap_exception(CapEx_TagViolation, auth_idx);
@@ -1731,7 +1729,7 @@ function clause execute (LoadResCapDDC(cd, rs1)) = {
   if haveAtomics() then {
     let ddc_val = DDC;
     let vaddr = ddc_val.address + X(rs1);
-    handle_loadres_cap_via_cap(cd, DDC_IDX, ddc_val, vaddr)
+    handle_loadres_cap_via_cap(cd, DDC_IDX, ddc_val, vaddr, false, false)
   } else {
     handle_illegal();
     RETIRE_FAIL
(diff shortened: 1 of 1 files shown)


rmn30:
Also worth noting that this is untested.

rmn30:
Hold fire, looks like I still have to do some work  on the handling of extended page table walk info for AMOSWAP.

Makefile
```diff
@@ -185,7 +185,11 @@ SOFTFLOAT_LIBS   = $(SOFTFLOAT_LIBDIR)/softfloat.a
 SOFTFLOAT_SPECIALIZE_TYPE = RISCV
 
 GMP_FLAGS = $(shell pkg-config --cflags gmp)
-GMP_LIBS = $(shell pkg-config --libs gmp)
+GMP_LIBS := $(shell pkg-config --libs gmp)
+ifeq (${GMP_LIBS},)
+# GMP does not have pkg-config metadata on Ubuntu 18.04 so default to this:
+GMP_LIBS := -lgmp
+endif
```
jrtc27, line 192:
```
# GMP does not have pkg-config metadata on Ubuntu 18.04 so provide a default:
GMP_LIBS := $(shell pkg-config --libs gmp || echo -lgmp)
```
maybe?

src/cheri_insts.sail
```diff
@@ -2021,7 +2033,7 @@ function handle_store_cond_cap_via_cap(cs2, auth_idx, auth_val, vaddrBits) = {
     RETIRE_FAIL
   } else if match_reservation(vaddrBits) == false then {
     /* cannot happen in rmem */
-    C(cs2) = int_to_cap(EXTZ(0b1));
+    X(rd) = EXTZ(0b1);
```
jrtc27, line 2036: Hmm, for explicit SC of a cap this was deliberately written using `C` so it works for split register files

Makefile
rmn30: Yes, that is more concise.

src/cheri_insts.sail
rmn30: That's a good point, I didn't think about that. Since it only ever writes an integer I thought it made sense to use the integer register file. In a split register file world the versions of the instructions that have a source / dest operand could read from a capability register and write to an integer register with the same number. That would avoid having to do a `cgetaddr` in order to branch on the return value but I guess it would be a nightmare for the compiler's register allocator?

Makefile
rmn30: Now done.

src/cheri_insts.sail
rmn30: I've now pushed a version where the destination register is either a capability or integer depending on the instruction.

PeterRugg:
Testing this in TestRIG, I came accross:

`Sail exception!Exiting due to uncaught exception: sail-riscv/model/riscv_mem.sail 108:31 - 108:68`

when running `.4byte 0x1201292f # lr.w x18, x2 (rl) .`

I remember this already being an issue in Sail: that rl and aq snag with some trap. It makes it a little difficult to test this PR: I guess I'll have to restrict to only test amos that don't have these bits set.

rmn30:
I will take a look at that. The assembly doesn't seem to indicate they are set?

jrtc27:
> I will take a look at that. The assembly doesn't seem to indicate they are set?

TestRIG's disassembler is full of quirks; see the `(rl)` after the instruction.

rmn30:
The RISC-V spec. says:
> Software should not set the *rl* bit on an LR instruction unless the *aq* bit is also set, nor should
software set the *aq* bit on an `SC` instruction unless the *rl* bit is also set. `LR.rl` and `SC.aq` instructions
are not guaranteed to provide any stronger ordering than those with both bits clear, but may result
in lower performance.

Apparently these instructions don't really make much sense (load release is not a thing) but it is not very helpful for the Sail model to crash. The problem is that we have to convert these to one of the [memory access types](https://github.com/rems-project/sail/blob/sail2/lib/instr_kinds.sail) understood by RMEM but none of them fits. For the sequential execution model it doesn't make any difference so we could arbitrarily pick `RISCV_plain` (downgrade) or `RISCV_strong_aquire` (upgrade?) but I am not sure.

Not that the AMO ops [logically AND the two bits together](https://github.com/rems-project/sail-riscv/blob/master/model/riscv_insts_aext.sail#L217) before calling read / write but I'm not sure if this is correct.

jrtc27:
> The RISC-V spec. says:
> 
> > Software should not set the _rl_ bit on an LR instruction unless the _aq_ bit is also set, nor should
> > software set the _aq_ bit on an `SC` instruction unless the _rl_ bit is also set. `LR.rl` and `SC.aq` instructions
> > are not guaranteed to provide any stronger ordering than those with both bits clear, but may result
> > in lower performance.
> 
> Apparently these instructions don't really make much sense (load release is not a thing) but it is not very helpful for the Sail model to crash. The problem is that we have to convert these to one of the [memory access types](https://github.com/rems-project/sail/blob/sail2/lib/instr_kinds.sail) understood by RMEM but none of them fits. For the sequential execution model it doesn't make any difference so we could arbitrarily pick `RISCV_plain` (downgrade) or `RISCV_strong_aquire` (upgrade?) but I am not sure.
> 
> Not that the AMO ops [logically AND the two bits together](https://github.com/rems-project/sail-riscv/blob/master/model/riscv_insts_aext.sail#L217) before calling read / write but I'm not sure if this is correct.

I think it's more sensible to downgrade them in the sail model. I think the warning is just there because simple hardware implementations might not look at the actual AMO itself and either (a) see some of the bits are set, so it'll do a full fence because that's all it has, or (b) unnecessarily do a read fence when storing, or unnecessarily do a write fence when loading.

rmn30:
Downgrade sounds like it would comply with spec. too, but would be nice to see a warning or something.

rmn30:
Note that this is dependent on sail-riscv PR which is not yet merged: https://github.com/rems-project/sail-riscv/pull/89

PR #48, opened 2021-06-10
Github action

rmn30:
This adds a github action that builds the C emulators for RV32 and RV64, normal and rvfi on each push. Unfortunately compiling sail and install its dependencies each time takes a while and is quite inefficient. We might be able to use a docker file to improve on this but I've not figured out how yet. Fortunately pushes are not that frequent.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -18,5 +18,5 @@ jobs:
       uses: actions/checkout@HEAD
       with:
         submodules: true
-    - name: Build model
-      run: eval $(opam env) && make -d csim
+    - name: Build simulators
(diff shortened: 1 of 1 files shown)


.github/workflows/compile.yml
```diff
@@ -0,0 +1,22 @@
...
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: [ubuntu-18.04]
+    steps:
+    - name: Install opam2
+      run: |
+        sudo add-apt-repository -y ppa:avsm/ppa
```
arichardson, line 11: Could also download `https://github.com/ocaml/opam/releases/download/2.0.8/opam-2.0.8-x86_64-linux` instead of using the ppa?

arichardson: Probably makes sense to check that it builds on both.

PR #51, opened 2021-09-29
WIP: Indirect sentries

nwf:
This sits on top of #18 at the moment, just because it needs to read capabilities from memory.


src/cheri_types.sail
```diff
@@ -60,9 +60,10 @@
 /*  SUCH DAMAGE.                                                                         */
 /*=======================================================================================*/
 
-let reserved_otypes = 4
+let reserved_otypes = 8
```
jrtc27, line 63: This can stay as 4

src/cheri_insts.sail
```diff
@@ -108,6 +110,48 @@ function clause execute(CJAL(imm, cd)) = {
...
+    match load_cap_via_cap(0b0 @ cs, cs_val, idcaddr) {
+      None() => RETIRE_FAIL,
+      Some(v) => {
+        if not (v.tag) then {
+          handle_cheri_reg_exception(CapEx_TagViolation, cs);
+          RETIRE_FAIL
+        } else if not (v.permit_execute) then {
+          handle_cheri_reg_exception(CapEx_PermitExecuteViolation, cs);
+          RETIRE_FAIL
+        } else {
```
jrtc27, line 139: Need to handle v being sealed (especially if it points to a sentry, which is a bit redundant but IMO should be allowed). At which point this basically looks like a copy of CJALR, except it rejects nested indirect sentries, so probably worth sharing code.

src/cheri_insts.sail
```diff
@@ -135,6 +179,9 @@ function clause execute(CJALR(imm, cs1, cd)) = {
   if not (cs1_val.tag) then {
     handle_cheri_reg_exception(CapEx_TagViolation, cs1);
     RETIRE_FAIL
+  } else if signed(cs1_val.otype) == otype_isentry_ptp then {
+    /* Indirect points-to-PCC sentries have their own requirements; jump out */
+    handle_cjalr_isentry_ptp(cd, cs1)
   } else if isCapSealed(cs1_val) &
             ((signed(cs1_val.otype) != otype_sentry) | imm != zeros()) then {
```
jrtc27, line 186: Immediate check needs to apply to indirect sentries too (as it would imply loading from a different address; currently you just silently ignore it). I do wonder though if CJALR should support these transparently or not, for Morello there are separate indirect and indirect-pair sentry branch instructions. I imagine not knowing whether or not a CJALR will need to load would be a complete pain for OoO processors as which execution unit it is issued to has a value dependency on its source register.

src/cheri_insts.sail
```diff
@@ -135,6 +179,9 @@ function clause execute(CJALR(imm, cs1, cd)) = {
   if not (cs1_val.tag) then {
     handle_cheri_reg_exception(CapEx_TagViolation, cs1);
     RETIRE_FAIL
+  } else if signed(cs1_val.otype) == otype_isentry_ptp then {
```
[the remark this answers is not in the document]
nwf: Erm, quite.  `ipcc`?

src/cheri_insts.sail
nwf: Thanks for the reminder; let's follow Morello on this and we can always experiment with making `cjalr` follow transparently later.

src/cheri_types.sail
```diff
@@ -63,6 +63,8 @@
 let reserved_otypes = 4
 let otype_unsealed = -1
 let otype_sentry   = -2 /* Sealed erstwhile or prospective PCC */
+let otype_isentry_load = -3 /* Sealed erstwhile IDC pointing at PCC */
```
arichardson, line 66: Is `erstwhile` correct? Isn't this what is going to be `idc` when invoked rather than what was idc when sealed?

nwf: How embarrassing.

src/cheri_insts.sail
```diff
@@ -1656,6 +1726,172 @@ function clause execute (CLoadTags(rd, cs1)) = {
...
+  } else if not (cs1_val.permit_load_cap) then {
+    handle_cheri_reg_exception(CapEx_PermitLoadCapViolation, cs1);
+    RETIRE_FAIL
+  } else {
+    C(cd) = sealCap(cs1_val, to_bits(cap_otype_width, otype_isentry_load));
+    RETIRE_SUCCESS
+  }
+}
+
+union clause ast = CInvokeLAL : (regidx, regidx)
```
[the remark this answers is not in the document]
nwf: I think the intention was "Load And Link", though I no longer remember.  It's a bad name.

PR #52, opened 2021-10-01
CBuildCap: reject sentries without permit_execute

arichardson:
We can't create those normally, so CBuildCap also shouldn't allow it.

### DIFF
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -916,7 +916,7 @@ function clause execute (CBuildCap(cd, cs1, cs2)) = {
       RETIRE_FAIL
     } else {
       assert(exact, "CBuildCap: setCapBounds was not exact"); /* base and top came from cs2 originally so will be exact */
-      let cd5 = if signed(cs2_val.otype) == otype_sentry then sealCap(cd4, to_bits(cap_otype_width, otype_sentry)) else cd4;
+      let cd5 = if signed(cs2_val.otype) == otype_sentry & cs2_val.permit_execute then sealCap(cd4, to_bits(cap_otype_width, otype_sentry)) else cd4;
       C(cd) = cd5;
       RETIRE_SUCCESS
     }


arichardson:
I found this local diff in my repo, I believe it's because QEMU has this check.

src/cheri_insts.sail
```diff
@@ -916,7 +916,7 @@ function clause execute (CBuildCap(cd, cs1, cs2)) = {
       RETIRE_FAIL
     } else {
       assert(exact, "CBuildCap: setCapBounds was not exact"); /* base and top came from cs2 originally so will be exact */
-      let cd5 = if signed(cs2_val.otype) == otype_sentry then sealCap(cd4, to_bits(cap_otype_width, otype_sentry)) else cd4;
+      let cd5 = if signed(cs2_val.otype) == otype_sentry & cs2_val.permit_execute then sealCap(cd4, to_bits(cap_otype_width, otype_sentry)) else cd4;
```
rmn30, line 919: Should this raise a permit execute violation?

rmn30: Also could do with some line breaks.

rmn30 (commented):
Agree that the check should be present but see comments.

src/cheri_insts.sail
arichardson: Possibly, I'm just matching https://github.com/CTSRD-CHERI/qemu/commit/cd35dc7102d54fd36b7761997318bcbfe9cd43cb here. Not sure what hardware does in this case.

rmn30: Toooba does not check for the executable permission on sentry caps. There is an `isDerivable` check but it only checks the exponent is in range, the reserved bits are zero and the top and base bits look sensible. Maybe it should be the job of `CCopyType` and `CCSeal` to restore special sealed capabilities like sentries and, in future, indirect sentries?

In an ideal world we would use the Sail to prove that `CBuildCap`, `CCopyType` and `CCSeal` can only be used to create capabilities that are derivable by other instructions and that any derivable instruction can be restored from bits :-)

jrtc27: Doing it with CCSeal does not make sense. CSeal cannot make sentries, and CCSeal is just a conditional version of it. Nor should they, there is no authorising cap. Having it be CBuildCap was a very deliberate choice.

It also means it's only one instruction for every function pointer you need to initialise rather than three.

jrtc27: And yeah, we should be throwing an exception here, leaving it unsealed but valid is a bad idea

PeterRugg: Just wondering what goes wrong if we give up enforcing this: hopefully all the sail code checks the permit_execute bit when jumping to sentries anyway.

nwf: See https://github.com/CTSRD-CHERI/cheri-architecture/issues/40 for the discussion we had last time about this.  I broadly agree that, until we find architectural purpose and a non-cbuildcap way of constructing such capabilities, that cbuildcap shouldn't construct them, but I think it would also be fine to clear the tag on the cbuildcap result for otype==sentry and !Permit_Execute rather than raising an exception.

jrtc27: In a tag-clearing world, yeah, you'd tag-clear here, but that's not what QEMU does, it just skips the sealing step...

nwf: Certainly skipping the sealing step is a bug, yes.

nwf:
I've folded this up into #51 since I need to touch the same place.  Please let me know if that (e.g., bc04ad177aa47e49a131679f46f8ef4aa7120504) variant looks OK?

jrtc27:
> I've folded this up into #51 since I need to touch the same place. Please let me know if that (e.g., [bc04ad1](https://github.com/CTSRD-CHERI/sail-cheri-riscv/commit/bc04ad177aa47e49a131679f46f8ef4aa7120504)) variant looks OK?

That has a dependency on revocation 3.0 PTE bits, and itself will need review. This consistency fix is self-contained and can go in today.

jrtc27:
Well, except this needs fixing to trap not silently create an unsealed cap...

nwf:
Well, maybe fetch bc04ad1 out from #51?

arichardson:
The better approach is to allow CSealEntry to also seal non-executable capabilities (this can be useful even if they can't be called). Obsoleted by #56

PR #54, opened 2021-10-27
Load/store as user

nwf:
Sits on top of the CLG and isentry PRs because orthogonality is hard.  The last two commits are the salient ones: some preparatory refactoring and then the actual instruction bodies.


nwf:
@PeterRugg Thanks for looking; I've responded in a `fixup!` commit, for ease of incremental review, that I'll squash before merging to mainline.

PR #56, opened 2022-02-09
Replace trapping behaviours with tag clearing where safe.

PeterRugg:
As discussed in https://github.com/CTSRD-CHERI/cheri-architecture/issues/80, move from trapping to tag clearing.

This PR goes all the way with the change, including sealing and unsealing, but I'm very happy to hear feedback on how far we should take it.

Some of the syntax could possibly be improved also. I've consolidated all the tag clearings into `invalidateCap` and `invalidateCapUnless` calls, but this could be a bad idea or maybe there's a neater way to do it.

Currently I haven't run any tests. I will be TestRIGging against the hardware implementations as I change them, but I'm keen to hear ideas about what else to run.

Have any properties been proved of CHERI RISC-V? (@bauereiss ?). It might be interesting to check they still pass after these changes.

I would suggest not merging until we check there aren't too many timing surprises in hardware, especially around CSetBounds.


src/cheri_insts.sail
```diff
@@ -950,29 +759,16 @@ union clause ast = CCopyType : (regidx, regidx, regidx)
...
-    handle_cheri_reg_exception(CapEx_TagViolation, cs1);
-    RETIRE_FAIL
-  } else if isCapSealed(cs1_val) then {
-    handle_cheri_reg_exception(CapEx_SealViolation, cs1);
-    RETIRE_FAIL
-  } else if hasReservedOType(cs2_val) then {
+  if hasReservedOType(cs2_val) then {
     C(cd) = int_to_cap(EXTS(cs2_val.otype));
     RETIRE_SUCCESS
-  } else if cs2_otype < cs1_base then {
```
rmn30, line 966: These checks have disappeared. I suppose they are subsumed by the less strict `invalidateCapUnless(newCap, representable)` below combined with the checks in `CCSeal`?

rmn30 (commented):
Looks good. I like these changes and refactoring make the sail more readable. I assume those familiar with compiler / sw implications already had input?

src/cheri_insts.sail
```diff
@@ -882,48 +718,29 @@ union clause ast = CBuildCap : (regidx, regidx, regidx)
...
+             & (cs2_base <= cs2_top) /* check for length < 0 - possible because cs2 might be untagged */
+             & ((cs2_perms & cs1_perms) == cs2_perms);
+  let (exact, cd1) = setCapBounds(cs1_val, to_bits(cap_addr_width, cs2_base), to_bits(cap_len_width, cs2_top));
+  let (_, cd2) = setCapOffset(cd1, getCapOffsetBits(cs2_val)); // Ignore representability check, since Fast Rep Check not relevant
+  let cd3 = setCapPerms(cd2, cs2_perms);
+  let cd4 = setCapFlags(cd3, cs2_flags);
+  let derivable = cd4 == {unsealCap(cs2_val) with tag=cd4.tag};
+  assert(not(derivable) | exact, "CBuildCap: setCapBounds was not exact"); /* base and top came from cs2 originally so will be exact */
+  let cd5 = if signed(cs2_val.otype) == otype_sentry then sealCap(cd4, to_bits(cap_otype_width, otype_sentry)) else cd4;
+  let cd6 = if subset & derivable then cd5 else invalidateCap(cs1_val);
```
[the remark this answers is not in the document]
PeterRugg: Hardware just wants to (maybe) set the tag. If the capability isn't derivable, it shouldn't have to replace it with the capability that gets derived from the extracted fields.

src/cheri_insts.sail
```diff
@@ -403,7 +400,7 @@ union clause ast = CSetFlags : (regidx, regidx, regidx)
  */
 function clause execute(CSetFlags(cd, cs1, rs2)) = {
   let cs1_val' = C(cs1);
-  let cs1_val = {cs1_val' with tag = cs1_val'.tag & not(isCapSealed(cs1_val'))};
+  let cs1_val = invalidateCapUnless(cs1_val', not(isCapSealed(cs1_val')));
```
arichardson, line 403: I personally find this easier to read as `invalidateCapIf(cs1_val', isCapSealed(cs1_val'))`. No strong preference either way though.

src/cheri_insts.sail
```diff
@@ -882,48 +718,29 @@ union clause ast = CBuildCap : (regidx, regidx, regidx)
...
+  let subset = (cs2_base >= cs1_base)
+             & (cs2_top <= cs1_top)
+             & (cs2_base <= cs2_top) /* check for length < 0 - possible because cs2 might be untagged */
+             & ((cs2_perms & cs1_perms) == cs2_perms);
+  let (exact, cd1) = setCapBounds(cs1_val, to_bits(cap_addr_width, cs2_base), to_bits(cap_len_width, cs2_top));
+  let (_, cd2) = setCapOffset(cd1, getCapOffsetBits(cs2_val)); // Ignore representability check, since Fast Rep Check not relevant
+  let cd3 = setCapPerms(cd2, cs2_perms);
+  let cd4 = setCapFlags(cd3, cs2_flags);
+  let derivable = cd4 == {unsealCap(cs2_val) with tag=cd4.tag};
+  assert(not(derivable) | exact, "CBuildCap: setCapBounds was not exact"); /* base and top came from cs2 originally so will be exact */
```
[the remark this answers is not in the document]
rmn30: OK, then I think we should change the comment to say something more useful like "if cs2 was a valid (derivable) capability encoding (i.e. no reserved bits set, sensible bounds bits) then the setBounds should be exact." By the way I think you are confusing `cs1` (authority) and `cs2` (template).

src/cheri_insts.sail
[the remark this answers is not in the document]
rmn30: OK, but then shouldn't it be `cs2_val` (the unmodified template) with tag unset?

src/cheri_insts.sail
rmn30: Agreed, the double negative is a bit confusing.

src/cheri_insts.sail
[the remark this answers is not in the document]
PeterRugg: Oh, hmm, even that's not quite right, because hardware will still want to clear the type. I'll do a fix.

src/cheri_insts.sail
```diff
@@ -726,7 +587,7 @@ union clause ast = CClearTag : (regidx, regidx)
  */
 function clause execute (CClearTag(cd, cs1)) = {
   let cs1_val = C(cs1);
-  C(cd) = {cs1_val with tag=false};
+  C(cd) = invalidateCap(cs1_val);
```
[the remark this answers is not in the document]
PeterRugg: Agreed. I changed `invalidateCap` to `clearTag`, since I think we're pretty settled on not legalizing to null. That's as explicit, but still consistent with all the other places where the function is used.

src/cheri_insts.sail
PeterRugg: Yep, I was thinking of it as "the properties that must be true after this for the tag to be set", but it just confused things. I've changed to `clearTagIf`, and negated (or unnegated) the conditions.

src/cheri_insts.sail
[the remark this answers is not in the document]
PeterRugg: Okay, I've clarified. I'm not sure where I've confused `cs1` with `cs2`, or maybe I already fixed that when factoring out `targetCap`?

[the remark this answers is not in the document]
ronorton: > the base and length weren't really what cs1 was asking for
I think I was only referring to this comment -- it is `cs2` doing the asking!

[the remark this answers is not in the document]
ronorton: It might be useful to rename `cs1_val` and `cs2_val` to make their functions clear.

src/cheri_insts.sail
ronorton: Sounds sensible!

src/cheri_insts.sail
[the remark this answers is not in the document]
PeterRugg: Done... Settled on `authority` and `requested` since they have the same length.

src/cheri_insts.sail
```diff
@@ -705,26 +705,26 @@ union clause ast = CBuildCap : (regidx, regidx, regidx)
...
-  let (exact, cd1) = setCapBounds(cs1_val, to_bits(cap_addr_width, cs2_base), to_bits(cap_len_width, cs2_top));
-  let (_, cd2) = setCapOffset(cd1, getCapOffsetBits(cs2_val)); // Ignore representability check, since Fast Rep Check not relevant
-  let cd3 = setCapPerms(cd2, cs2_perms);
-  let cd4 = setCapFlags(cd3, cs2_flags);
-  let derivable = cd4 == targetCap;
-  assert(not(derivable) | exact, "CBuildCap: setCapBounds was not exact"); /* base and top came from cs2 originally so will be exact */
-  let cd5 = if signed(cs2_val.otype) == otype_sentry then sealCap(cd4, to_bits(cap_otype_width, otype_sentry)) else cd4;
-  let cd6 = if subset & derivable then cd5 else clearTag(targetCap);
+  let authority_val = clearTagIfSealed(if unsigned(cs1) == 0 then DDC else C(cs1));
+  let requested_val = {C(cs2) with tag=authority_val.tag};
```
PeterRugg, line 709: This allows you to `CBuildCap` an already tagged cap. It might clear the tag if the `CBuildCap` would fail. Seems harmless to me, but I don't know if others have thoughts.

nwf: That does seem a little strange, though it's hard to see a case where we'd knowingly be passing a tagged cap as the bag of bits to `CBuildCap` and possibly clearing the tag doesn't seem like it will make such an unforeseen situation any worse than it already would be.

That said, if we're changing things, might we prefer shifting to `CCBuildCap`-esque behaviour where tagged things are just passed through unmodified (for non-branching reconstruction with multiple roots).

src/cheri_cap_common.sail
```diff
@@ -466,7 +466,14 @@ function getCapCursor(cap) : Capability -> CapAddrInt = unsigned(cap.address)
...
 
-function invalidateCap(cap) : Capability -> Capability = {cap with tag=false}
+function clearTagIf(cap, cond) : (Capability, bool) -> Capability =
+    {cap with tag = cap.tag & not(cond)}
+
+function clearTagIfSealed(cap) : Capability -> Capability =
+    clearTagIf(cap, isCapSealed(cap))
+
+function clearTag(cap) : Capability -> Capability =
+    clearTagIf(cap, true)
```
arichardson, line 476: Not sure if this extra indirection makes the code more readable, I'd just go for
```suggestion
    {cap with tag = false}
```

src/cheri_insts.sail
```diff
@@ -882,48 +705,29 @@ union clause ast = CBuildCap : (regidx, regidx, regidx)
...
+  let (requested_base, requested_top) = getCapBounds(requested_val);
+  let authority_perms = getCapPerms(authority_val);
+  let requested_perms = getCapPerms(requested_val);
+  let requested_flags = getCapFlags(requested_val);
+  let subset = (requested_base >= authority_base)
+             & (requested_top <= authority_top)
+             & (requested_base <= requested_top) /* check for length < 0 - possible because requested might be untagged */
+             & ((requested_perms & authority_perms) == requested_perms);
+  let (exact, cd1) = setCapBounds(authority_val, to_bits(cap_addr_width, requested_base), to_bits(cap_len_width, requested_top));
+  let (_, cd2) = setCapOffset(cd1, getCapOffsetBits(requested_val)); // Ignore representability check, since Fast Rep Check not relevant
```
arichardson, line 722: Inconsistent use of `//` and  `/*` comments here.

arichardson (commented):
LGTM with history cleaned up. I don't have an opinion either way on the seal/unseal changes, so I'm fine to go with tag clearing.

src/cheri_insts.sail
```diff
@@ -647,76 +534,39 @@ union clause ast = CSetBoundsImmediate : (regidx, regidx, bits(12))
...
-  } else if isCapSealed(cs1_val) then {
-    handle_cheri_reg_exception(CapEx_SealViolation, cs1);
-    RETIRE_FAIL
-  } else if not(inCapBounds(cs1_val, newBase, unsigned(rs2_val))) then {
-    handle_cheri_reg_exception(CapEx_LengthViolation, cs1);
-    RETIRE_FAIL
-  } else {
-    let (exact, newCap) = setCapBounds(cs1_val, newBase, newTop);
-    if not (exact) then {
-      handle_cheri_reg_exception(CapEx_InexactBounds, cs1);
```
rwatson, line 713: Does this mean that CapEx_InexactBounds can now be garbage collected?

rmn30: Probably. Also CapEx_UserDefViolation, CapEx_PermitSealViolation, CapEx_PermitUnsealViolation and maybe CapEx_TypeViolation?

src/cheri_insts.sail
```diff
@@ -1054,53 +836,23 @@ union clause ast = CSeal : (regidx, regidx, regidx)
...
-    handle_cheri_reg_exception(CapEx_TagViolation, cs2);
-    RETIRE_FAIL
-  } else if isCapSealed(cs1_val) then {
-    handle_cheri_reg_exception(CapEx_SealViolation, cs1);
-    RETIRE_FAIL
-  } else if isCapSealed(cs2_val) then {
-    handle_cheri_reg_exception(CapEx_SealViolation, cs2);
-    RETIRE_FAIL
-  } else if not (cs2_val.permit_seal) then {
-    handle_cheri_reg_exception(CapEx_PermitSealViolation, cs2);
```
rwatson, line 1089: Possibly can also GC CapEx_PermitSealViolation?

src/cheri_insts.sail
```diff
@@ -1190,87 +918,37 @@ union clause ast = CUnseal : (regidx, regidx, regidx)
...
-    handle_cheri_reg_exception(CapEx_SealViolation, cs2);
-    RETIRE_FAIL
-  } else if hasReservedOType(cs1_val) then {
-    handle_cheri_reg_exception(CapEx_TypeViolation, cs1);
-    RETIRE_FAIL
-  } else if cs2_cursor != unsigned(cs1_val.otype) then {
-    handle_cheri_reg_exception(CapEx_TypeViolation, cs2);
-    RETIRE_FAIL
-  } else if not (cs2_val.permit_unseal) then {
-    handle_cheri_reg_exception(CapEx_PermitUnsealViolation, cs2);
```
rwatson, line 1232: And could perhaps GC CapEx_PermitUnsealViolation?

src/cheri_insts.sail
bsdjhb: AccessCInvokeIDC has already been removed from the spec it was MIPS-only from when CCall had a delay slot.  CallTrap, ReturnTrap, and TSSUnderFlow are for the old CCall selector 0 and are no longer in the spec either.

PeterRugg: Oh, I see they're gone in master. I'll do a rebase.

src/cheri_insts.sail
```diff
@@ -974,54 +933,26 @@ union clause ast = CUnseal : (regidx, regidx, regidx)
...
- *   - *cs2*.**tag** is not set.
- *   - *cs1* is not sealed.
- *   - *cs2* is sealed.
- *   - *cs1*.**otype** is reserved.
- *   - *cs2*.**address** $\ne$ *cs1*.**otype**.
- *   - *cs2*.**perms** does not grant **Permit_Unseal**.
- *   - *cs2*.**address** $\lt$ *cs2*.**base**.
- *   - *cs2*.**address** $\ge$ *cs2*.**top**.
+ * is stripped of **Global**. If *cs2* is unable to authorize the unsealing,
+ * the **tag** field of *cd* is cleared.
```
nwf, line 922: Almost surely worth a comment giving justification?  Something like... "We believe it to be simpler and more flexible for compartmentalized software to check the tag of the result of unsealing an input capability rather than immediately trapping if the otype is out of range.  If software omits such a check and directly uses the result of unsealing for memory dereference, a malformed input will cause a tag-clear trap on the dereference instead."  Probably needs some workshopping.

src/cheri_insts.sail
```diff
@@ -850,46 +850,24 @@ union clause ast = CSeal : (regidx, regidx, regidx)
...
- *
- * An exception is raised if:
- *   - *cs2*.**tag** is not set.
- *   - *cs2*.**sealed** is set.
- *   - *cs2*.**perms** does not grant **Permit_Seal**.
- *   - *cs2*.**address** $\lt$ *cs2*.**base**.
- *   - *cs2*.**address** $\ge$ *cs2*.**top**.
- *   - *cs2*.**address** $\gt$ [cap_max_otype].
+ * *cs2*. If *cs2* is unable to authorize the sealing, the **tag** field of
+ * *cd* is cleared.
```
nwf, line 840: I suppose the justification here is by symmetry with `CUnseal` (below) and `CCSeal` (which intrinsically has kind of unusual pass-through behavior)... probably worth a slightly expanded comment?

src/cheri_insts.sail
```diff
@@ -2013,8 +1998,8 @@ function clause execute AMOSwapCap(cd, cs2, rs1_cs1, aq, rl) = {
                   match wres {
                     MemValue(_) => {
                       let cr = if ptw_info.ptw_lc == PTW_LC_CLEAR
-                        then invalidateCap(v) /* strip the tag */
-                        else invalidateCapUnless(v, auth_val.permit_load_cap);
+                        then clearTag(v) /* strip the tag */
+                        else clearTagIf(v, not(auth_val.permit_load_cap));
```
nwf, line 2003: Nit: perhaps simpler to just flatten this to
`let cr = clearTagIf(v, (ptw_info.ptw_lc == PTW_LC_CLEAR) | not(auth_val.permit_load_cap));`?

nwf (commented):
Some minor nits but generally LGTM.

src/cheri_insts.sail
```diff
@@ -438,35 +411,19 @@ union clause ast = CToPtr : (regidx, regidx, regidx)
...
-    RETIRE_FAIL
-  } else {
-    /* Note: returning zero for untagged values breaks magic constants such as SIG_IGN */
-    X(rd) = if not (cs1_val.tag) then
-              zeros()
-            else
-              cs1_val.address - getCapBaseBits(cs2_val);
-    RETIRE_SUCCESS
-  }
+  let cs2_val = if unsigned(cs2) == 0 then DDC else C(cs2);
```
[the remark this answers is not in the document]
PeterRugg: Removed by @jrtc27 [here](https://github.com/CTSRD-CHERI/sail-cheri-riscv/commit/5bb148625ddcc3cc406b0d3cb842453547b2cfab). I think I'm mildly in favour of it being explicit, but open to changing it if people prefer.

src/cheri_insts.sail
PeterRugg: Do you mean inline in the comment and latex, or in the rationalisation section? I imagine the ISA doc will grow a detailed section on tag-clearing vs trapping decisions (initial draft lives [here](https://github.com/CTSRD-CHERI/cheri-specification/pull/1) that needs expansion, especially to talk about sealing). Comments in the sail are otherwise very rationale-light, and I guess we'd need this comment on Seal, Unseal, and all variant friends?

src/cheri_insts.sail
```diff
@@ -647,76 +534,39 @@ union clause ast = CSetBoundsImmediate : (regidx, regidx, bits(12))
...
- *
- * An exception is raised if:
- *   - *cs1*.**tag** is not set.
- *   - *cs1* is sealed.
- *   - *cs1*.**address** $\lt$ *cs1*.**base**.
- *   - *cs1*.**address** $+$ *rs2* $\gt$ *cs1*.**top**.
- *   - The resulting capability cannot be represented exactly.
+ * replaced with integer register *rs2*. If the resulting capability cannot be
+ * represented exactly, the **tag** field will be cleared (unlike [CSetBounds])
+ * , the **base** will be rounded down and the **length** will be rounded up by
```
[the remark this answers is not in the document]
PeterRugg: Well spotted, and you're quite right. Fixed to be less awful.

src/cheri_insts.sail
```diff
@@ -859,21 +693,10 @@ union clause ast = CBuildCap : (regidx, regidx, regidx)
...
- *   - *cs1* is sealed.
- *   - *cs2*.**base** $\lt$ *cs1*.**base**
- *   - *cs2*.**top** $\gt$ *cs1*.**top**
- *   - *cs2*.**base** $\gt$ *cs2*.**top**
- *   - *cs2*.**perms** grants a permission that is not granted by
- *     *cs1*.**perms**.
- *   - *cs2*.**uperms** grants a permission that is not granted by
- *     *cs1*.**uperms**.
- *
+ * *cs2* is a sentry then *cd* is also sealed as a sentry. If *cd* is not a
```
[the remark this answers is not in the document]
PeterRugg: Yep, a misreading could lead to a very evil implementation. Fixed to be in line with other similar comments.

src/cheri_insts.sail
PeterRugg: Yeah, I think I'd left it this way for symmetry with the cases that still needed the `if` to do reservation things, but you're right that this is clearer.

src/cheri_insts.sail
nwf: Yeah, on second thought, you're right; the rationale belongs elsewhere in the ISA doc rather than in the more mechanically-minded sail comments.

src/cheri_insts.sail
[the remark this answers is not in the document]
nwf: Nah, is fine as is.

nwf (approved):

src/cheri_insts.sail
```diff
@@ -733,31 +733,25 @@ function clause execute (CBuildCap(cd, cs1, cs2)) = {
...
-    C(cd) = int_to_cap(EXTS(cs2_val.otype));
-    RETIRE_SUCCESS
-  } else {
-    let (representable, newCap) = setCapAddr(cs1_val, to_bits(cap_addr_width, cs2_otype));
-    C(cd) = clearTagIf(newCap, not(representable));
-    RETIRE_SUCCESS
-  }
+  let reserved = hasReservedOType(cs2_val);
+  let otype : xlenbits = if reserved then EXTS(cs2_val.otype)
+                                     else EXTZ(cs2_val.otype);
```
nwf, line 751: Oh, that makes me kind of nervous... at least a big comment about all reserved otypes being "negative" right now?

jrtc27: In what sense? You'll find the same code in CGetType.

src/cheri_insts.sail
```diff
@@ -882,99 +706,52 @@ union clause ast = CBuildCap : (regidx, regidx, regidx)
...
-      handle_cheri_reg_exception(CapEx_LengthViolation, cs2);
-      RETIRE_FAIL
-    } else {
-      assert(exact, "CBuildCap: setCapBounds was not exact"); /* base and top came from cs2 originally so will be exact */
-      let cd5 = if signed(cs2_val.otype) == otype_sentry then sealCap(cd4, to_bits(cap_otype_width, otype_sentry)) else cd4;
-      C(cd) = cd5;
-      RETIRE_SUCCESS
-    }
-  }
+  let authority_val = clearTagIfSealed(if unsigned(cs1) == 0 then DDC else C(cs1));
```
jrtc27, line 709: This is a bit of a wall of code. Some blank lines to separate reading registers from checks from the actual operations might go a long way to help that.

jrtc27 (commented):
Descriptions need updating in a number of places to document the tag-clearing behaviour, otherwise they imply that certain invalid operations (e.g. CSetAddr on a sealed capability with a representable address) give you back a capability with the tag still set.

src/cheri_insts.sail
```diff
@@ -383,53 +380,29 @@ union clause ast = CAndPerm : (regidx, regidx, regidx)
...
- *
- * ## Exceptions
- *
- * An exception is raised if:
- *   - *cs1*.**tag** is not set.
- *   - *cs1* is sealed.
  */
 function clause execute(CAndPerm(cd, cs1, rs2)) = {
-  let cs1_val = C(cs1);
+  let cs1_val = clearTagIfSealed(C(cs1));
```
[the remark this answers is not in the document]
PeterRugg: Hmm, yeah, it's a shame because I didn't want to hide away the fact that it's the register value you're working on... How would you feel about `input_val` or `input_cap` for the result of maybe clearing the tag?

src/cheri_insts.sail
```diff
@@ -438,35 +411,19 @@ union clause ast = CToPtr : (regidx, regidx, regidx)
...
-    X(rd) = if not (cs1_val.tag) then
-              zeros()
-            else
-              cs1_val.address - getCapBaseBits(cs2_val);
-    RETIRE_SUCCESS
-  }
+  let cs2_val = if unsigned(cs2) == 0 then DDC else C(cs2);
+  let cs1_val = clearTagIfSealed(C(cs1));
+  /* Note: returning zero for untagged values breaks magic constants such as SIG_IGN */
+  X(rd) = if not (cs1_val.tag & cs2_val.tag) then
```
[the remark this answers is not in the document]
PeterRugg: Hmm, yes. I figured if the original was worried enough about it to throw a trap in this case then we'd want to signal the error somehow. I have no strong feelings though. Are you confident that not checking the tag is the way to go, or do you know if anyone else is likely to have thoughts?

src/cheri_insts.sail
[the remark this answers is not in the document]
PeterRugg: Okay, I've avoided this idiom. Does this look any better?

src/cheri_insts.sail
PeterRugg: Hopefully sorted, and roughly in keeping with the new whitespace in other non-trivial instructions

src/cheri_insts.sail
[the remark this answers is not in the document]
PeterRugg: Changed to match Morello

src/cheri_insts.sail
```diff
@@ -438,35 +419,22 @@ union clause ast = CToPtr : (regidx, regidx, regidx)
...
-  } else {
-    /* Note: returning zero for untagged values breaks magic constants such as SIG_IGN */
-    X(rd) = if not (cs1_val.tag) then
-              zeros()
-            else
-              cs1_val.address - getCapBaseBits(cs2_val);
-    RETIRE_SUCCESS
-  }
+
+  let inCap = clearTagIfSealed(cs1_val);
```
jrtc27, line 430: Not needed

src/cheri_insts.sail
```diff
@@ -383,53 +380,37 @@ union clause ast = CAndPerm : (regidx, regidx, regidx)
...
-    let perms = getCapPerms(cs1_val);
-    let mask = truncate(rs2_val, cap_perms_width);
-    let newCap = setCapPerms(cs1_val, (perms & mask));
-    C(cd) = newCap;
-    RETIRE_SUCCESS
-  }
+
+  let inCap = clearTagIfSealed(cs1_val);
+  let perms = getCapPerms(inCap);
+  let mask = truncate(rs2_val, cap_perms_width);
```
jrtc27, line 390: I feel it'd be better to getCapPerms on cs1_val and do the clearTagIfSealed with the setCapPerms (whether inline or just moving the let)?

src/cheri_insts.sail
```diff
@@ -383,53 +380,37 @@ union clause ast = CAndPerm : (regidx, regidx, regidx)
...
-    handle_cheri_reg_exception(CapEx_SealViolation, cs1);
-    RETIRE_FAIL
-  } else {
-    let newCap = setCapFlags(cs1_val, truncate(rs2_val, cap_flags_width));
-    C(cd) = newCap;
-    RETIRE_SUCCESS
-  }
+
+  let inCap = clearTagIfSealed(cs1_val);
+
```
jrtc27, line 409: I wouldn't include this blank line (repeated a bunch)

src/cheri_insts.sail
```diff
@@ -608,36 +541,22 @@ union clause ast = CSetBounds : (regidx, regidx, regidx)
...
- *   - *cs1*.**address** $+$ *rs2* $\gt$ *cs1*.**top**.
+ * capability covering the requested bounds. The **tag** field of the result
+ * is cleared if the bounds of the result exceed the bounds of *cs1*.
  */
 function clause execute (CSetBounds(cd, cs1, rs2)) = {
   let cs1_val = C(cs1);
   let rs2_val = X(rs2);
-  let newBase = cs1_val.address;
+
+  let inCap = clearTagIfSealed(cs1_val);
```
jrtc27, line 551: As with CAndPerm

src/cheri_insts.sail
```diff
@@ -647,76 +566,47 @@ union clause ast = CSetBoundsImmediate : (regidx, regidx, bits(12))
...
- *   - *cs1*.**address** $\lt$ *cs1*.**base**.
- *   - *cs1*.**address** $+$ *uimm* $\gt$ *cs1*.**top**.
+ * the requested bounds. The **tag** field of the result is cleared if the
+ * bounds of the result exceed the bounds of *cs1*.
  */
 function clause execute (CSetBoundsImmediate(cd, cs1, uimm)) = {
   let cs1_val = C(cs1);
-  let newBase = cs1_val.address;
+
+  let inCap = clearTagIfSealed(cs1_val);
```
jrtc27, line 575: Ditto

src/cheri_insts.sail
```diff
@@ -647,76 +566,47 @@ union clause ast = CSetBoundsImmediate : (regidx, regidx, bits(12))
...
+ * rounded up by the smallest amount needed to form a representable capability
+ * covering the requested bounds. The **tag** field of the result is cleared
+ * if the bounds of the result exceed the bounds of *cs1*.
  */
 function clause execute (CSetBoundsExact(cd, cs1, rs2)) = {
   let cs1_val = C(cs1);
   let rs2_val = X(rs2);
-  let newBase = cs1_val.address;
+
+  let inCap = clearTagIfSealed(cs1_val);
```
jrtc27, line 601: Ditto

src/cheri_insts.sail
```diff
@@ -824,32 +714,19 @@ union clause ast = CFromPtr : (regidx, regidx, regidx)
...
- *   - *cs1*.**tag** is not set.
- *   - *cs1* is sealed.
  */
 function clause execute (CFromPtr(cd, cs1, rs2)) = {
-  let (cs1_val, cs1_idx) = if unsigned(cs1) == 0 then (DDC, DDC_IDX)
-                                                 else (C(cs1), 0b0 @ cs1);
+  let cs1_val = if unsigned(cs1) == 0 then DDC else C(cs1);
   let rs2_val = X(rs2);
+
+  let inCap = clearTagIfSealed(cs1_val);
```
jrtc27, line 722: Given comments elsewhere I might be tempted to push this into the else

src/cheri_insts.sail
```diff
@@ -882,99 +749,63 @@ union clause ast = CBuildCap : (regidx, regidx, regidx)
...
-      RETIRE_FAIL
-    } else {
-      assert(exact, "CBuildCap: setCapBounds was not exact"); /* base and top came from cs2 originally so will be exact */
-      let cd5 = if signed(cs2_val.otype) == otype_sentry then sealCap(cd4, to_bits(cap_otype_width, otype_sentry)) else cd4;
-      C(cd) = cd5;
-      RETIRE_SUCCESS
-    }
-  }
+
+  let authorityCap = clearTagIfSealed(cs1_val);
```
jrtc27, line 755: The authority is cs1_val, not this

src/cheri_insts.sail
```diff
@@ -882,99 +749,63 @@ union clause ast = CBuildCap : (regidx, regidx, regidx)
...
-    handle_cheri_reg_exception(CapEx_LengthViolation, cs1);
-    RETIRE_FAIL
-  } else {
-    let (success, cap) = setCapAddr(cs1_val, to_bits(cap_addr_width, cs2_otype));
-    assert(success, "CopyType: offset is in bounds so should be representable");
-    C(cd) = cap;
-    RETIRE_SUCCESS
-  }
+
+  let authCap = clearTagIfSealed(cs1_val);
```
jrtc27, line 800: As with CBuildCap, not an authority

Would also move to next to the setCapAddr

src/cheri_insts.sail
```diff
@@ -1054,53 +888,28 @@ union clause ast = CSeal : (regidx, regidx, regidx)
...
- *   - *cs2*.**address** $\ge$ *cs2*.**top**.
- *   - *cs2*.**address** $\gt$ [cap_max_otype].
+ * *cs2*. If *cs2* is unable to authorize the sealing, the **tag** field of
+ * *cd* is cleared.
  */
 function clause execute (CSeal(cd, cs1, cs2)) = {
   let cs1_val = C(cs1);
   let cs2_val = C(cs2);
+
+  let authCap = clearTagIfSealed(cs1_val);
```
jrtc27, line 898: Ditto

src/cheri_insts.sail
```diff
@@ -1498,9 +1235,8 @@ function handle_load_cap_via_cap(cd, auth_idx, auth_val, vaddrBits) = {
       let c = mem_read_cap(addr, aq, aq & rl, false);
       match c {
         MemValue(v) => {
-          let cr = if ptw_info.ptw_lc == PTW_LC_CLEAR
-                   then {v with tag = false} /* strip the tag */
-                   else {v with tag = v.tag & auth_val.permit_load_cap};
+          let cr = clearTagIf(v, ptw_info.ptw_lc == PTW_LC_CLEAR
+                               | not(auth_val.permit_load_cap));
```
jrtc27, line 1239: Odd indentation

src/cheri_insts.sail
```diff
@@ -2320,9 +2056,8 @@ function clause execute AMOSwapCap(cd, cs2, rs1_cs1, aq, rl) = {
                   let wres : MemoryOpResult(bool) = mem_write_cap(addr, cs2_val, aq & rl, rl, false);
                   match wres {
                     MemValue(_) => {
-                      let cr = if ptw_info.ptw_lc == PTW_LC_CLEAR
-                        then {v with tag = false} /* strip the tag */
-                        else {v with tag = v.tag & auth_val.permit_load_cap};
+                      let cr = clearTagIf(v, ptw_info.ptw_lc == PTW_LC_CLEAR
+                                           | not(auth_val.permit_load_cap));
```
jrtc27, line 2060: Ditto

src/cheri_insts.sail
PeterRugg: Hmm, I thought of the "clearTagIfSealed" as like a guard that stops any authority leaking into the semantics, which is why initially I put it with the register read, so I was inclined to use `inCap` as much as possible. I guess you want to just put it on the main path through the function, which I can see. I'll give it a go and see how it looks.

jrtc27: Yeah, my view is it's logically on the "I'm mutating the capability now" output path, not on the input path

src/cheri_insts.sail
PeterRugg: Hmm, it does change the behaviour because of the inCap.tag case: if the authority was sealed do we want to give zeros or just give the pointer?

jrtc27: That's true... the two should really go together. Morello takes the same approach and just doesn't care, 0 is based solely on the being-converted capability, not the authority, where only the base matters (if at all, since turning DDCBO off makes it a CGetAddr for the non-zero case).

PR #57, opened 2022-07-21
Fix a comment about where legalize_epcc is used.

rmn30:
This is quite important as it affects the 
behaviour of sentries installed in xEPCC.


nwf:
Well that's just embarrassing.  The comment was correct when it was introduced in https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/21/commits/3565d5e10b75eaf9ce1fc557b528bade41204064, but within the same PR, https://github.com/CTSRD-CHERI/sail-cheri-riscv/pull/21/commits/09250510b11d752f0b2ef3c7fda9a516bc35ff1 made it out of date.  D'oh!

nwf (approved):

PR #59, opened 2022-08-24
Add CGetTop

arichardson:
This instruction returns the top of a capability (i.e. one byte past the
end), or an all-ones value for full-address-space capabilities (just
like CGetLen). This can be used instead of a saturating add of base+length.


nwf (approved):
LGTM

arichardson:
@jrtc27 does this look okay?

arichardson:
ping

PR #61, opened 2022-10-31
Add bool for writing to destination reg to RVFI output

francislaus:
In sail, when an instruction writes to a register, the register number will also be written to the `rvfi_rd_addr` field (if it has been built for RVFI output). When an instruction has no destination register, the `rvfi_rd_addr` remains zero. The CClear instruction in the CHERI is an exception to the regular RISC-V instruction format. It can write to up to eight registers. However, CClear has no destination register, but rather a two-bit immediate determining which quarter of the register file this instruction will write to.

Currently, the sail implementation for CClear writes the highest register number to `rvfi_rd_addr`. This PR changes this behaviour and never writes a destination register to the field, which results in the field containing zeros. I think this is the correct behaviour for the model as the `rvfi_rd_addr` field cannot correctly represent the destination register behaviour of CClear.

### DIFF
--- a/src/cheri_insts.sail
+++ b/src/cheri_insts.sail
@@ -783,7 +783,7 @@ function clause execute (CClear(q, m)) = {
       if q_u == 0 & i == 0 then
         DDC = null_cap
       else
-        C(8 * q_u + i) = null_cap;
+        wC (false, 8 * q_u + i, null_cap);
   RETIRE_SUCCESS
 }
 
--- a/src/cheri_regs.sail
+++ b/src/cheri_regs.sail
@@ -103,8 +103,8 @@ function rC r = {
 }
 
 /* writes a register with a capability value */
-val wC : forall 'n, 0 <= 'n < 32. (regno('n), regtype) -> unit effect {wreg, escape}
-function wC (r, v) = {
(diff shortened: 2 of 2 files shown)


src/cheri_insts.sail
```diff
@@ -783,7 +783,7 @@ function clause execute (CClear(q, m)) = {
       if q_u == 0 & i == 0 then
         DDC = null_cap
       else
-        C(8 * q_u + i) = null_cap;
+        wC (false, 8 * q_u + i, null_cap);
```
PeterRugg, line 786: This is the bit that will show in the ISA doc, and the no-context `false` may be confusing. I wonder if there's a clearer way, for example wrapping wc(false, ....) in a function. Not sure what to call that function though, since RVFI isn't really visible from the instruction-level description.

PeterRugg:
> Why not have it auto detect when it’s been called twice?

I thought about that, but what if you call it with an immediate where only one bit is set?

jrtc27:
Then it reports a single register write, which is accurate?

PeterRugg:
Hmm, I guess that's one way of looking at it. It might be a little tricky to have Toooba detect that case, although if we're not hoping to synthesise with RVFI there's probably an easy enough hacky way to do it.

PR #62, opened 2022-11-12
Add an implementation of the proposed CRelocate instruction

jrtc27:


PeterRugg (approved):

src/cheri_cap_common.sail
```diff
@@ -549,14 +549,66 @@ function capToString (cap) = {
...
+  getBoundsAlignmentMask(c)
+}
+
 function getRepresentableLength(len) : xlenbits -> xlenbits = {
   let m = getRepresentableAlignmentMask(len);
   (len + ~(m)) & m
 }
+
+function relocateCap(cap, offset) : (Capability, CapAddrBits) -> (bool, Capability) = {
+  let E = min(cap_max_E, unsigned(cap.E));
```
arichardson, line 570: This is a rather large and complex piece of code that looks like a close duplicate of other decoding code. I think this could either do with comments and/or share more code with getCapBoundsBits.

arichardson (commented):
Commit message should probably reference https://github.com/CTSRD-CHERI/cheri-specification/issues/3?

PR #64, opened 2023-01-16
Don't print CHERI exceptions unconditionally

arichardson:
Guard this with get_config_print_platform() to match other exception information print calls.

### DIFF
--- a/src/cheri_addr_checks.sail
+++ b/src/cheri_addr_checks.sail
@@ -66,7 +66,7 @@ val handle_cheri_cap_exception : (CapEx, capreg_idx) -> unit effect {escape, rre
 function handle_cheri_cap_exception(capEx, regnum) =
   {
     if get_config_print_platform()
-    then print("CHERI " ^ string_of_capex(capEx) ^ " Reg=" ^ BitStr(regnum));
+    then print_platform("CHERI " ^ string_of_capex(capEx) ^ " Reg=" ^ BitStr(regnum));
     let t : sync_exception = struct {
       trap    = E_Extension(EXC_CHERI),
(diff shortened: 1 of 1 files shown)


arichardson:
ping?

rmn30:
Hi, sorry for not responding previously. I agree that this message should not be printed unconditionally. I do find it useful to see CHERI exceptions printed though so locally I have some changes to
1. add `--trace=exception` to `riscv_sim.c`
2. make this predicated on `config_print_instr | config_print_exception`. This means you can either get a verbose instruciton trace with `-V --trace=instr` or just exceptions with `-V --trace=exeption`. I used `instr` as the predicate but `platform` might make more sense?
3. I'm considering changing the default to be non-verbose in `riscv_sim.c`.

arichardson:
Sounds good to me, I just noticed this message since it was being printed a lot while fuzzing and there is no way to turn it off. Happy to wait until your changes land.

arichardson:
@rmn30 are you planning to work on this? If not I'd like to place this under config_print_platform until a better tracing solution has been implemented.

rmn30:
I already did this for the CHERIoT model. Unfortunately the exact commit was squashed when we open sourced but you can see the [diff here](https://github.com/microsoft/cheriot-sail/commit/85757532bf65cc98ce4754ddeb559b8c654222f8#diff-03ce501b8766837b9d7c57e1caab15204fa1dcfdc330005bd7d642d366a33147L68). It depends on a patch to [sail-riscv](https://github.com/microsoft/cheriot-sail/blob/main/riscv_patches/0006-Add-a-new-exception-trace-option-that-prints-only-on.patch) so it's probably best to just merge your change.

PR #65, opened 2023-01-16
Remove unnecessary let in get_cheri_mode_cap_addr

arichardson:
I saw this slighly odd construct while reading through the file and it appears that sail accepts the code even without the let statement. My guess is that this was required in older versions of sail?

CC: @Alasdair

### DIFF
--- a/src/cheri_addr_checks.sail
+++ b/src/cheri_addr_checks.sail
@@ -174,8 +174,7 @@ function get_cheri_mode_cap_addr (base_reg : regidx, offset : xlenbits) = {
     let base_cap = C(base_reg) in
     (base_cap, base_cap.address + offset, 0b0 @ base_reg)
   else
-    let ddc = DDC in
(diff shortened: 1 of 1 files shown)


rmn30:
If DDC is a register then I think the new code will end up generating two reads from DDC. It's probably fine in emulation but might not be what you want in the monadic models.

arichardson:
> If DDC is a register then I think the new code will end up generating two reads from DDC. It's probably fine in emulation but might not be what you want in the monadic models.

Ah thanks that makes sense, for some reason I assumed this was a sail limitation.

rmn30:
If it confused you then it might be worth a comment to make it clearer. Maybe a better name for the let bound would help?

arichardson:
> If it confused you then it might be worth a comment to make it clearer. Maybe a better name for the let bound would help?

Yeah I guess changing the name or a comment would help, but it's not that important.

PR #66, opened 2023-01-19
Update to use new capability mode decoding hook

arichardson:
This depends on the upstream pull request https://github.com/riscv/sail-riscv/pull/205. The new approach is more flexible compared to the previous ast remapping one since we can return completely different results rather than having to match the same arguments. This will be required to implement compressed CHERI instruction decoding where some of them have different immmediates compared to the baseline ISA.

Additionally, this fixes the assembly tracing output for capability-mode instructions such as auipcc (which was printed as auipc before).


arichardson:
rebased now that the upstream change has been merged.

src/cheri_insts.sail
```diff
@@ -2358,3 +2358,30 @@ mapping clause encdec = LOADRES(aq, rl, rs1, size, rd) if (size == BYTE) | (size
...
+scattered mapping encdec_capmode
+
+mapping clause encdec_capmode = AUIPCC(imm, cd) <-> imm @ cd @ 0b0010111
+mapping clause assembly = AUIPCC(imm, cd)
+  <-> "auipcc" ^ spc() ^ cap_reg_name(cd) ^ sep() ^ hex_bits_20(imm)
+mapping clause encdec_capmode = CJALR(imm, cs1, cd) <-> imm @ cs1 @ 0b000 @ cd @ 0b1100111
+mapping clause assembly = CJALR(imm, cs1, cd)
+  <-> "cjalr" ^ spc() ^ cap_reg_name(cd) ^ sep() ^ cap_reg_name(cs1) ^ sep() ^ hex_bits_12(imm)
+/* See `mapping clause encdec = RISCV_JAL` for why we have to use this syntax */
+mapping clause encdec_capmode = CJAL(imm_19 @ imm_7_0 @ imm_8 @ imm_18_13 @ imm_12_9 @ 0b0, cd)
```
arichardson, line 2377: I find the naming of these immediate operands somewhat confusing (I'd prefer them to use immediate bitpos index rather than encoding bitpos) but this matches the base model.

arichardson:
ping? This is needed for any future PRs that update the sail-riscv submodule.

src/cheri_decode_ext.sail
```diff
@@ -60,13 +60,16 @@
 /*  SUCH DAMAGE.                                                                         */
 /*=======================================================================================*/
 
+val ext_decode_compressed : bits(16) -> ast effect {rreg}
```
arichardson, line 63: Could drop this effect if we are happy with moving to sail 0.15

arichardson:
Removed effects. Should hopefully be okay to merge?

PeterRugg (approved):

PR #67, opened 2023-01-20
Sync default instructions list with sail-riscv

arichardson:
This does not yet include any of the Zb* or Zk* instructions, but adds various compressed instructions (e.g. hints) that I noticed were missing while fuzzing against QEMU.

Counter-example trace:
```
.4byte 0x00000062 # SLLI hint
#  A < Trap:  True, PCWD: 0x0000000000000000, RD: 00, RWD: 0x0000000000000000, MA: 0x0000000000000000, MWD: 0x0000000000000000, MWM: 0b00000000, I: 0x0000000000000069 PRV_M XL:64 (Unknown instruction)
#  B > Trap: False, PCWD: 0x0000000080000002, RD: 00, RWD: 0x0000000000000000, MA: 0x0000000000000000, MWD: 0x0000000000000000, MWM: 0b00000000, I: 0x0000000000000069 PRV_M XL:64 (Unknown instruction)
.assert rd_wdata == 0x0 "hint"
```


nwf (approved):
Seems OK by me

PR #68, opened 2023-01-31
Consistently use `not(` instead of `not (`

arichardson:
There were 67 occurrences of `not(` vs 31 for `not (`, so I changed those to the more common style. The base RISC-V model doesn't define `not` and uses `~` instead, so maybe we should also use that style?


arichardson:
I find `not()` a bit more readable for boolean conditions than `~`, but happy to change it to `~` if that's preferred.

jrtc27:
This is definitely an improvement. I'm also torn; `not()` is more readable, but `~()` is the standard RISC-V model function...

arichardson:
I'm going to go ahead an merge this, we can always globally replace `not()` with `~()` if we decide to match the base model.

PR #69, opened 2023-02-03
Add support for compressed CHERI instructions

arichardson:


arichardson:
Missing c.jal (RV32 only)

arichardson:
Should be ready for review now - I believe I've add all currently defined instructions.

rmn30:
This looks like a fine approach. It's a bit sad that the encdec mappings have to be duplicated from the RISC-V versions. The approach I took for [CHERIoT](https://github.com/microsoft/cheriot-sail/blob/main/src/cheri_decode_ext.sail) is much more concise because it makes use of the existing RISC-V decoding mappings but the new way is better because it gets the disassembly correct. I agree that it's clearer to shift the immediate during decode, although interestingly it makes the mapping clause non-total in the reverse direction (encode). I don't think that's an issue -- encode is not really used and I think there are other cases.

I trust you've got the details of all the encodings correct. I'll try merging it to CHERIoT and run our tests then we can be fairly confident it's right.

arichardson:
I wonder if these should be defined in a separate `cheri_insts_cext.sail` or are we happy having everything in one file?

rmn30:
One argument for moving them to a separate file is that if you put `scattered mapping` and `end mapping` in different files it is easy to add new clauses to the mapping for new instructions by adding new files providing you everhing to Sail in the correct order. Not exactly elegant but it works and is about as extensible as Sail gets currently. This is how the `encdec` and `assembly` mappings work in sail-riscv: see https://github.com/riscv/sail-riscv/blob/master/model/riscv_insts_begin.sail and https://github.com/riscv/sail-riscv/blob/master/model/riscv_insts_end.sail .

src/cheri_insts_cext.sail
```diff
@@ -0,0 +1,170 @@
...
+/*=======================================================================================*/
+
+/* Compressed capability loads/stores (encodings differ between RV32/RV64) */
+union clause ast = C_CLC : (bits(9), cregidx, cregidx)
+
+mapping clause encdec_compressed_capmode = C_CLC(ui8 @ ui76 @ ui54 @ 0b0000, cs1c, cdc) if sizeof(xlen) == 64
+  <-> 0b001 @ ui54 : bits(2) @ ui8 : bits(1) @ cs1c : cregidx @ ui76 : bits(2) @ cdc : cregidx @ 0b00 if sizeof(xlen) == 64
+mapping clause encdec_compressed_capmode = C_CLC(0b0 @ ui76 @ ui53 @ 0b000, cs1c, cdc) if sizeof(xlen) == 32
+  <-> 0b011 @ ui53 : bits(3) @ cs1c : cregidx @ ui76 : bits(2) @ cdc : cregidx @ 0b00 if sizeof(xlen) == 32
+mapping clause assembly = C_CLC(uimm, cdc, csc1)
```
rmn30, line 70: `cdc` and `csc1` swapped here vs decode clauses. Same in assembly clause.

Also inconsistent use of `cs1c` vs `csc1`.

rmn30 (changes requested):
Looks good except for the bug above.

src/cheri_insts_cext.sail
rmn30: This fix works but means the order of the operands is now 'middle-endian' matching neither the encoding or the `LoadCapImm` constructor. I think it would be slightly less confusing to use the same order as `LoadCapImm`.

rmn30: Well, I suppose it matches the encoding as far as it can it's just unfortunately that `LoadCapImm` chose a different order.

src/cheri_insts_cext.sail
```diff
@@ -0,0 +1,170 @@
...
+}
+
+union clause ast = C_CIncOffset4CSPN : (cregidx, bits(10))
+
+mapping clause encdec_compressed_capmode = C_CIncOffset4CSPN(cdc, nz96 @ nz54 @ nz3 @ nz2 @ 0b00)
+      if nz96 @ nz54 @ nz3 @ nz2 != 0b00000000
+  <-> 0b000 @ nz54 : bits(2) @ nz96 : bits(4) @ nz2 : bits(1) @ nz3 : bits(1) @ cdc : cregidx @ 0b00
+      if nz96 @ nz54 @ nz3 @ nz2 != 0b00000000
+mapping clause assembly = C_CIncOffset4CSPN(cdc, nzuimm) if nzuimm != 0b0000000000
+  <-> "c.cinoffet4cspn" ^ spc() ^ cap_creg_name(cdc) ^ sep() ^ hex_bits_10(nzuimm) if nzuimm != 0b0000000000
```
rmn30, line 137:
```suggestion
  <-> "c.cincoffet4cspn" ^ spc() ^ cap_creg_name(cdc) ^ sep() ^ hex_bits_10(nzuimm) if nzuimm != 0b0000000000
```

rmn30: There was a missing 'c' and 's' in 'incoffset'. Not sure if the 'n' is significant?

arichardson: The n matches `c.addi4spn`

arichardson:
Updated to match base instruction operand order.

jrtc27 (commented):
I *believe* this will incorrectly permit `c.flwsp f0, imm(reg)` on RV32 and `c.fldsp f0, imm(reg)` on RV64 rather than making them illegal to match `c.l[wd]sp`? You need both "illegal" and "fallback on normal RVC" return values. This is a bit of an annoying special-case.

src/cheri_insts_end.sail
```diff
@@ -0,0 +1,75 @@
...
+/*  ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,                   */
+/*  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT                   */
+/*  OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF                   */
+/*  SUCH DAMAGE.                                                                         */
+/*=======================================================================================*/
+
+/* Put the illegal instructions last to use their wildcard match. */
+
+/* Wilcard match for 32-bit non-capmode encodings. */
+union clause ast = CAPMODE_ILLEGAL : word
```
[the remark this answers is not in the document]
rmn30: I had the same thought but not sure what would be better.

rmn30:
Would like to see some comments explaining which RVC instruction each encoding replaces.

src/cheri_insts_end.sail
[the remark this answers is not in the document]
arichardson: I guess I could use `NOT_CAPMODE`?

src/cheri_insts_cext.sail
```diff
@@ -0,0 +1,170 @@
...
+/*  ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,                   */
+/*  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT                   */
+/*  OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF                   */
+/*  SUCH DAMAGE.                                                                         */
+/*=======================================================================================*/
+
+/* Compressed capability loads/stores (encodings differ between RV32/RV64) */
+union clause ast = C_CLC : (bits(9), cregidx, cregidx)
+
+mapping clause encdec_compressed_capmode = C_CLC(ui8 @ ui76 @ ui54 @ 0b0000, cdc, csc1) if sizeof(xlen) == 64
```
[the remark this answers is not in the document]
arichardson: That's what I originally had since I copy-pasted those. But IMO it's nicer this way since the assembly output is actually useful and we can avoid doing the extending in the execute clause. This makes the functions that differ between 32/64 nicer.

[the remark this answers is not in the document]
arichardson: Oh you mean `creg2reg_idx(cdc)` in the encdec clause? Wasn't sure if that would work.

arichardson:
Added documentation comments and hopefully addressed the raised issues.

arichardson:
> I _believe_ this will incorrectly permit `c.flwsp f0, imm(reg)` on RV32 and `c.fldsp f0, imm(reg)` on RV64 rather than making them illegal to match `c.l[wd]sp`? You need both "illegal" and "fallback on normal RVC" return values. This is a bit of an annoying special-case.

I am not sure how this would happen? In RV32 capmode `c.flwsp f0, imm(reg)` will match `c.cscsp`, and in non-capmode it will fall back to calling encdec_compressed which will return `C_ILLEGAL`?

src/cheri_insts_end.sail
```diff
@@ -0,0 +1,75 @@
...
+/*  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT                   */
+/*  OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF                   */
+/*  SUCH DAMAGE.                                                                         */
+/*=======================================================================================*/
+
+/* Put the illegal instructions last to use their wildcard match. */
+
+/* Wilcard match for 32-bit non-capmode encodings. */
+union clause ast = NOT_CAPMODE : word
+function clause execute NOT_CAPMODE(s) = { internal_error("should be unreachable"); RETIRE_FAIL }
```
rmn30, line 67: Minor thing, but would be good if this error was more specific:
```suggestion
function clause execute NOT_CAPMODE(s) = { internal_error("should be unreachable execute(NOT_CAPMODE(" ^ BitStr(s) ^ "))"); RETIRE_FAIL }
```

src/cheri_insts_end.sail
```diff
@@ -0,0 +1,75 @@
...
+mapping clause encdec_capmode = NOT_CAPMODE(s) <-> s
+end encdec_capmode
+
+/* Wilcard match for 16-bit non-capmode encodings. */
+union clause ast = NOT_C_CAPMODE : half
+function clause execute NOT_C_CAPMODE(s) = { internal_error("should be unreachable"); RETIRE_FAIL }
```
rmn30, line 73: Same for this error. Just thinking of similar feedback on PR in sail-riscv repo.
```suggestion
function clause execute NOT_C_CAPMODE(s) = { internal_error("should be unreachable execute(NOT_C_CAPMODE(" ^ BitStr(s) ^ "))"); RETIRE_FAIL }
```

rmn30: LGTM

arichardson:
@rmn30 I believe I've fixed your oustanding issues. @jrtc27 Are you okay with this version?

jrtc27:
> > I _believe_ this will incorrectly permit `c.flwsp f0, imm(reg)` on RV32 and `c.fldsp f0, imm(reg)` on RV64 rather than making them illegal to match `c.l[wd]sp`? You need both "illegal" and "fallback on normal RVC" return values. This is a bit of an annoying special-case.
> 
> I am not sure how this would happen? In RV32 capmode `c.flwsp f0, imm(reg)` will match `c.cscsp`, and in non-capmode it will fall back to calling encdec_compressed which will return `C_ILLEGAL`?

Which is the problem. `c.cscsp c0, imm(reg)` is not a legal instruction, just as `c.s[wd]sp x0, imm(reg)` is not legal in non-capmode.

arichardson:
Ah I misparsed your previous comment as "it will decode the cap versions as fp". I see now that due to the != zero restriction we do fall back to flw/fld. I guess the easiest solution is adding a new illegal decoding pattern.

arichardson:
ping @jrtc27 @rmn30. Would really like to get this merged soon to unblock follow-up work.

jrtc27:
Why introduce a new union member rather than just using C_ILLEGAL?

arichardson:
> Why introduce a new union member rather than just using?

I tried that initially, but `C_ILLEGAL` is defined in `riscv_insts_end.sail` so it's not visible there yet.

Should there be an upstream change to move the AST clause to the start just keep the encdec in the end file? However, in my experience upstream changes take at least 3 weeks to land so I'd prefer to keep this approach for now and then fix it if we decide to change upstream?

jrtc27:
> > Why introduce a new union member rather than just using?
> 
> I tried that initially, but `C_ILLEGAL` is defined in `riscv_insts_end.sail` so it's not visible there yet.
> 
> Should there be an upstream change to move the AST clause to the start just keep the encdec in the end file? However, in my experience upstream changes take at least 3 weeks to land so I'd prefer to keep this approach for now and then fix it if we decide to change upstream?

I guess that's ok provided there's a TODO/XXX comment above it and it looks and behaves like C_ILLEGAL rather than C_CLCSP_RESERVED_CD0/c.clcsp.reserved.cd0 or whatever it is; having the ast clause in riscv_insts_end.sail is just an oversight, not realising that it could be useful to have available.

arichardson:
> > > Why introduce a new union member rather than just using?
> > 
> > 
> > I tried that initially, but `C_ILLEGAL` is defined in `riscv_insts_end.sail` so it's not visible there yet.
> > Should there be an upstream change to move the AST clause to the start just keep the encdec in the end file? However, in my experience upstream changes take at least 3 weeks to land so I'd prefer to keep this approach for now and then fix it if we decide to change upstream?
> 
> I guess that's ok provided there's a TODO/XXX comment above it and it looks and behaves like C_ILLEGAL rather than C_CLCSP_RESERVED_CD0/c.clcsp.reserved.cd0 or whatever it is; having the ast clause in riscv_insts_end.sail is just an oversight, not realising that it could be useful to have available.

Done. Unless there are any further comments, I'll merge this later this week.

arichardson:
> > > > Why introduce a new union member rather than just using?
> > > 
> > > 
> > > I tried that initially, but `C_ILLEGAL` is defined in `riscv_insts_end.sail` so it's not visible there yet.
> > > Should there be an upstream change to move the AST clause to the start just keep the encdec in the end file? However, in my experience upstream changes take at least 3 weeks to land so I'd prefer to keep this approach for now and then fix it if we decide to change upstream?
> > 
> > 
> > I guess that's ok provided there's a TODO/XXX comment above it and it looks and behaves like C_ILLEGAL rather than C_CLCSP_RESERVED_CD0/c.clcsp.reserved.cd0 or whatever it is; having the ast clause in riscv_insts_end.sail is just an oversight, not realising that it could be useful to have available.
> 
> Done. Unless there are any further comments, I'll merge this later this week.

Sent https://github.com/riscv/sail-riscv/pull/223 upstream.
Diff once that is merged:
```diff
diff --git a/src/cheri_insts_cext.sail b/src/cheri_insts_cext.sail
index 1b1ca52..2fa3af8 100644
--- a/src/cheri_insts_cext.sail
+++ b/src/cheri_insts_cext.sail
@@ -110,17 +110,11 @@ function clause execute(C_CLCSP(cd, uimm)) =
  * The c.clcsp encoding with cd == 0 should raise an illegal instruction trap
  * instead of falling back to the non-capmode (legal) instruction
  * `c.flwsp/fldsp f0` so we need an explicit capmode illegal match clause.
- * TODO: Should use C_ILLEGAL instead once this is exported from sail-riscv.
  */
-union clause ast = C_CLCSP_RESERVED_CD0 : (bits(10))
-mapping clause encdec_compressed_capmode = C_CLCSP_RESERVED_CD0(ui96 @ ui5 @ ui4 @ 0b0000) if sizeof(xlen) == 64
+mapping clause encdec_compressed_capmode = C_ILLEGAL(0b001 @ ui5 @ 0b00000 @ ui4 @ ui96 @ 0b10) if sizeof(xlen) == 64
   <-> 0b001 @ ui5 : bits(1) @ 0b00000 @ ui4 : bits(1) @ ui96 : bits(4) @ 0b10 if sizeof(xlen) == 64
-mapping clause encdec_compressed_capmode = C_CLCSP_RESERVED_CD0(0b0 @ ui86 @ ui5 @ ui43 @ 0b000) if sizeof(xlen) == 32
+mapping clause encdec_compressed_capmode = C_ILLEGAL(0b011 @ ui5 @ 0b00000 @ ui43 @ ui86 @ 0b10) if sizeof(xlen) == 32
```

rmn30 (approved):
Sorry I've been away. Looks good to me. Good spot @jrtc27!

PR #71, opened 2023-03-08
Fix sail/opam warnings

arichardson:


arichardson:
The only remaining warnings are in the base model now:
```
Warning: Deprecated sail-riscv/model/prelude.sail:78.22-43:
78 |val string_startswith = "string_startswith" : (string, string) -> bool
   |                      ^-------------------^
   | 
All external bindings should be marked as either monadic or pure

Warning: Deprecated sail-riscv/model/prelude.sail:88.57-70:
88 |val "reg_deref" : forall ('a : Type). register('a) -> 'a effect {rreg}
   |                                                         ^-----------^
   | 
Explicit effect annotations are deprecated. They are no longer used and can be removed.

Warning: Redundant type annotation on assignment to w_pte sail-riscv/model/riscv_vmem_sv39.sail:233.2-18:
233 |           w_pte : SV39_PTE = update_Ext(w_pte, ext);
    |  ^--------------^
    | 
Type is already known

Warning: Redundant type annotation on assignment to w_pte sail-riscv/model/riscv_vmem_sv48.sail:197.5-21:
197 |       w_pte : SV48_PTE = update_Ext(w_pte, ext);
    |     ^--------------^
    | 
Type is already known
```

rmn30 (approved):
Looks good.

PR #74, opened 2023-06-15
Allocate a feature bit in xCCSR to report tag-clearing semantics

arichardson:
This can be used by CheriBSD to probe the underlying CHERI semantics while there are still multiple incompatible versions. In the longer term, this register is unlikely to be part of a standardized RISC-V extension, so allocating bits for temporary transitions should be fine.

### DIFF
--- a/src/cheri_sys_regs.sail
+++ b/src/cheri_sys_regs.sail
@@ -65,6 +65,8 @@
 /* Capability control csr */
 
 bitfield ccsr : xlenbits = {
+  /* Bits allocated from 31 downwards are used for feature flags. */
+  tc      :  31,     /* tag-clearing error semantics */
   d       :  1,      /* dirty  */
   e       :  0       /* enable */
 }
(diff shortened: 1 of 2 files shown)


src/cheri_sys_regs.sail
```diff
@@ -65,6 +65,8 @@
 /* Capability control csr */
 
 bitfield ccsr : xlenbits = {
+  /* Bits allocated from 31 downwards are used for feature flags. */
+  tc      :  31,     /* tag-clearing error semantics */
```
arichardson, line 69: I am planning to add bit 30 for "DDC relocation disabled" here as I don't believe we want that to be writable as it is for Morello.

bsdjhb: That seems fine to me.

src/cheri_sys_regs.sail
```diff
@@ -65,6 +65,8 @@
...
   d       :  1,      /* dirty  */
```
bsdjhb, line 70: Do we also want to remove 'd' and 'e'?  I don't think CheriBSD has ever used them.  Not sure if we have any other software that ever has.

bsdjhb (approved):

src/cheri_sys_regs.sail
arichardson: That sounds good to me, although we should keep e until we have an implementation of the new semantics.

PR #76, opened 2023-06-16
Disable DDC relocation by default

arichardson:
DDC relocation is no longer a mandatory part of the spec and is highly unlikely to be included as part of an initial standardized version of CHERI-RISC-V. In case we do decide to add DDC relocation as an optional extensions, this commit uses a (currently) hardcoded boolean and a helper function (maybe_ddc_relocated_addr) to allow turning it back on again rather than removing the code outright.

See https://github.com/CTSRD-CHERI/cheri-specification/issues/47

### DIFF
--- a/src/cheri_sys_regs.sail
+++ b/src/cheri_sys_regs.sail
@@ -67,6 +67,7 @@
 bitfield ccsr : xlenbits = {
   /* Bits allocated from 31 downwards are used for feature flags. */
   tc      :  31,     /* tag-clearing error semantics */
+  ndr     :  30,     /* no DDC relocation */
   e       :  0       /* enable */
 }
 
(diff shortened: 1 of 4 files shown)


src/cheri_sys_regs.sail
```diff
@@ -84,6 +85,7 @@ function legalize_ccsr(c : ccsr, v : xlenbits) -> ccsr = {
   let v = Mk_ccsr(v);
   /* For now these bits are not really supported so hardwired to true */
```
bsdjhb, line 86: This comment only applied to d/e FWIW and now only applies to e.  I wonder if we want a separate comment before the feature bits just so this older comment doesn't seem to apply to the feature bits?

bsdjhb (approved):

davidchisnall (approved):

src/cheri_sys_regs.sail
arichardson: Should be fixed now.

bsdjhb:
I think the helper function is useful.

PR #77, opened 2023-06-20
Disable PCC relocation by default

arichardson:
This is somewhat more invasive than the change to disable DDC relocation, so we might want to remove the fallback code outright rather than keeping a have_pcc_relocation() configuration option.

### DIFF
--- a/src/cheri_pc_access.sail
+++ b/src/cheri_pc_access.sail
@@ -64,10 +64,11 @@
 
 val get_arch_pc : unit -> xlenbits effect {rreg}
 function get_arch_pc () =
-  PC - getCapBaseBits(PCC)
+  if have_pcc_relocation() then PC - getCapBaseBits(PCC) else PC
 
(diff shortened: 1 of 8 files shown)


src/cheri_pc_access.sail
```diff
@@ -64,7 +64,7 @@
 
 val get_arch_pc : unit -> xlenbits effect {rreg}
 function get_arch_pc () =
-  PC - getCapBaseBits(PCC)
+  if have_pcc_relocation() then PC - getCapBaseBits(PCC) else PC
 
 /* FIXME: The asymmetry between get_next_pc and set_next_pc where one
```
bsdjhb, line 69: Do you want to update this FIXME comment to say it only applies if have_pcc_relocation() is true?

arichardson: Updated

PeterRugg (approved):

PR #79, opened 2023-06-23
Move bounds check after PC/PCC misalignment checks

PeterRugg:
This allows hardware to delay the bounds check. See discussion at https://github.com/CTSRD-CHERI/cheri-specification/issues/54


arichardson (commented):
Seems fine to me - I can rebase the PCC relocation change after this is merged.

PR #82, opened 2023-08-07
Fix LaTeX documentation for ddc_and_result_addr

arichardson:
It needs to be on the `val`, adding it to the `function` prevents it from being visible in the CHERI specification.

### DIFF
--- a/src/cheri_addr_checks.sail
+++ b/src/cheri_addr_checks.sail
@@ -167,7 +167,8 @@ type ext_data_addr_error = (CapEx, capreg_idx)
  * Returns the current value of DDC (for subsequent access checks) as well as
  * the resulting address for a non-CHERI integer-based memory access.
  */
-function ddc_and_resulting_addr (addr : xlenbits) -> (Capability, xlenbits) = {
+val ddc_and_resulting_addr : (xlenbits) -> (Capability, xlenbits)
+function ddc_and_resulting_addr (addr) = {
   let ddc_val = DDC in
   (ddc_val, if (have_ddc_relocation()) then ddc_val.address + addr else addr)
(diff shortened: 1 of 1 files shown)


bsdjhb (approved):

bsdjhb:
Hmm, this did not fix the issue for me in the generated tex for the spec.  The only change in the tex sources after this was:

```
--- a/sail_latex_riscv/fnzddc_and_resulting_addr0f0e2d5c6123a71667b35ed17c44868e.tex
+++ b/sail_latex_riscv/fnzddc_and_resulting_addr0f0e2d5c6123a71667b35ed17c44868e.tex
@@ -1,4 +1,4 @@
-function #\hyperref[sailRISCVzddczyandzyresultingzyaddr]{ddc\_and\_resulting\_addr}# (addr : xlenbits) -> (Capability, xlenbits) = {
+function #\hyperref[sailRISCVzddczyandzyresultingzyaddr]{ddc\_and\_resulting\_addr}# (addr) = {
   let ddc_val = DDC in
   (ddc_val, if (#\hyperref[sailRISCVzhavezyddczyrelocation]{have\_ddc\_relocation}#()) then ddc_val.address + addr else addr)
 }
```

src/cheri_addr_checks.sail
```diff
@@ -167,7 +167,8 @@ type ext_data_addr_error = (CapEx, capreg_idx)
  * Returns the current value of DDC (for subsequent access checks) as well as
```
bsdjhb, line 167: I needed to change the line above this from `/**` to `/*!` and then it worked.

bsdjhb (approved):

PR #86, opened 2023-08-14
Fixups to allow Coq output to build

bacam:
These shouldn't affect any other targets; the most invasive changes are to the Makefile, but they just mirror the changes in the sail-riscv repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -360,7 +360,7 @@ endif
 
 generated_definitions/lem/$(ARCH)/riscv.lem: $(SAIL_SRCS) Makefile handwritten_support/dummy_assembly_mappings.sail
 	mkdir -p generated_definitions/lem/$(ARCH) generated_definitions/isabelle/$(ARCH)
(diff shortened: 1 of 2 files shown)


PeterRugg:
Apologies this got missed! Happy to merge if it's still relevant.

bacam:
I'm fixing this up, but it's hit a few minor bugs in Sail.  I'll probably finish it tomorrow.

PeterRugg:
Thanks so much!

bacam:
Is the `sail-riscv` submodule a cheri-specific branch?  A bugfix to Sail has revealed a small problem in the memory interface that's no longer present upstream.  It would be nice to fix it directly rather than override it in `sail-cheri-riscv`.

PeterRugg:
Yes: it's pointing at the "TestRIG" branch here: https://github.com/CTSRD-CHERI/sail-riscv/tree/testrig. Feel free to make PRs to that.

